PR #2, opened 2023-01-04
Code Review

SinghxRaj:


HabitLogger.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
TheCSharpAcademy, line 1: Could you please move the entire project to a folder named SinghxRaj.HabitTracker as per the code review article? 😁
image
https://www.thecsharpacademy.com/article/52

HabitLogger/Program.cs
```diff
@@ -0,0 +1,215 @@
...
+
+    while (string.IsNullOrEmpty(input) || !int.TryParse(input, out id))
+    {
+        Console.Write("Invalid input. Type a valid id: ");
+        input = Console.ReadLine();
+    }
+    return id;
+}
+
+static void AddNewLog(HabitLoggerConnection connection)
```
TheCSharpAcademy, line 101: It would be good to let the user add the date for the log as well. Let's say I didn't have access to my computer for a day. I wouldn't be able to create a log for that day. What do you think?

HabitLoggerLibrary/HabitLoggerConnection.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            int id = reader.GetInt32(0);
+            string date = reader.GetDateTime(1).ToString().Split(" ")[0];
+            int quantity = reader.GetInt32(2);
+            logs.Add(Tuple.Create(id, date, quantity));
+        }
+        return logs;
+    }
+
+    // Deletes a record in the database.
```
TheCSharpAcademy, line 75: I know it's tempting to comment a lot when learning, but try to explain your code via good naming of members (methods, variable, etc). Well-written code is almost entirely self-explanatory. The less you rely on comments the more you force yourself to name things carefully (which you do, by the way)

TheCSharpAcademy (changes requested):

HabitLogger.sln
SinghxRaj: Alright, I moved all the entire project into a folder named SinghxRaj.HabitTracker.

HabitLogger/Program.cs
SinghxRaj: It makes sense to allow users the ability to provide their date. It more flexible for the user! At the same time, I believe my approach of just logging the current date is a good in the sense that it's easier for users to add a new log since they don't need to add their date. This makes it quicker and easier for the users. At the end though, both approaches have their pros and cons and are good!

I updated the add new log function so that the user will have to provide a date.

HabitLoggerLibrary/HabitLoggerConnection.cs
SinghxRaj: I removed the redundant comments.

TheCSharpAcademy (approved):

PR #11, opened 2023-03-23
Habit tracker project

LucianoNicolasArrieta:


TheCSharpAcademy:
@LucianoNicolasArrieta Excellent work! I'm approving the project after initial review. 

I'd just like to add a couple of pointers for future projects: 

When you ask for the id for an operation, I recommend you show the records first, otherwise the user will be flying blind: 

image

Also, when use 'using' statements to manage connections, you don't need to close the connection, 'using' deals with it for you behind the scenes, as per this discussion: https://codereview.stackexchange.com/questions/273292/try-catch-finally-with-using-in-ado-net

image

PR #16, opened 2023-04-06
Habit Tracker attempt

jwhitt3r:
Create the Habit Tracker exercise focused on using the tutorial provided: https://www.youtube.com/watch?v=d1JIJdDVFjs.


TheCSharpAcademy:
@jwhitt3r Hey! I'm approving the project now but I just have one comment. When deleting or updating a record if I input an Id that doesn't exist I don't get an error message, I simply get the input prompt again. It's better user experience if you always prompt a message with what went wrong 😊

PR #17, opened 2023-04-10
complete

Speedierone:


HabitTracker.Speedierone/HabitTracker.Speedierone/Program.cs
```diff
@@ -0,0 +1,242 @@
...
+                    $"INSERT INTO drinking_water(date, quantity) VALUES('{date}', {quantity})";
+
+            tableCmd.ExecuteNonQuery();
+            connection.Close();
+        }
+    }
+
+    internal static string GetDateInput()
+    {
+        Console.WriteLine("Please enter date in format dd-mm-yyyy");
```
TheCSharpAcademy, line 93: Looks like your parser is checking for dd-MM-yy but you're asking the user to enter dd-mm-yyyy, could you please fix that? 😁

TheCSharpAcademy (changes requested):

HabitTracker.Speedierone/HabitTracker.Speedierone/Program.cs
Speedierone: Heya, I have corrected this now and pushed it. I think it should have updated let me know if it hasnt.

TheCSharpAcademy:
I'll approve the project, but just bear in mind that using en-US is breaking it for me.. Being in Australia I was still getting an error. I had to use this instead: 

image

Speedierone:
no problem i will keep an eye on that. To be honest im from the UK so that maybe why i was coming up with some areas as i believe the US use mm-dd-yy.

PR #24, opened 2023-05-04
Merging in Habit Tracker and both challenges

csm-stough:


TheCSharpAcademy (changes requested):
I'm very impressed with your work! You seem to have very good command of C#. Well done 😁

I'm going to request one small modification though. In terms of the user experience, there should be a way for the users to easily visualize the value per habit per day. At the moment I can only see the date:

image

csm-stough:
Ahh your right, I was going through and updating those record labels and
must have missed that one. Thanks! I'll fix that now. 😀

PR #25, opened 2023-05-05
Code review and commit

VocalNight:


TheCSharpAcademy (changes requested):
Could you please double check if the update option is working correctly? I'm getting this: 

image

VocalNight:
Fixed! Will need to create a new database.

PR #29, opened 2023-06-14
First Pull Request for Review

CWolffles:


HabitTracker.Wolffles/Habit.cs
```diff
@@ -0,0 +1,90 @@
+﻿using Microsoft.Data.Sqlite;
+
+namespace habit_tracker;
+class Habit : IHabit
+{
+	private string m_connectionString = "";
```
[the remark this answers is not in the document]
CWolffles: Thank you for all the reviewing! It's a good learning experience. 

I do have a question. I have my Habit class being called through a IHabit interface. Would getters and setters work through an interface? I can't seem to get it working.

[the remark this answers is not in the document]
kimfom01: No I don’t think so. It is meant for class members that hold data not methods. That’s why it doesn’t work for the interface methods

HabitTracker.Wolffles/HabitMenu.cs
```diff
@@ -0,0 +1,107 @@
...
+	{
+		Console.Clear();
+
+		do
+		{
+			Console.WriteLine($"Enter a new record. Enter number of {m_habit.GetMeasureUnit()}: ");
+		}
+		while (!m_habit.Insert(InputValidation.GetUserInputAsInt(), InputValidation.GetUserInputAsDate()));
+
+		Console.ReadKey();
```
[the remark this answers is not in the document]
CWolffles: Updated as requested. Good suggestion :)

HabitTracker.Wolffles/HabitMenu.cs
```diff
@@ -0,0 +1,107 @@
...
+	}
+	private void UpdateRecord()
+	{
+		Console.Clear();
+		m_habit.Read();
+
+		Console.WriteLine("Select a record to update by typing it's DATE");
+		string date = InputValidation.GetUserInputAsDate();
+
+		Console.WriteLine($"{date} selected. Enter value:");
```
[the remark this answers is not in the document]
CWolffles: Looks like the function wasn't being called correctly from by the HabitMenu class. Resolved on next Commit.

HabitTracker.Wolffles/Habit.cs
```diff
@@ -3,10 +3,10 @@
 namespace habit_tracker;
 class Habit : IHabit
 {
-	private string m_connectionString = "";
-	private string m_habitName = "";
-	private string m_habitTable = "";
-	private string m_habitUnit = "";
+	public string m_connectionString { get; set; }
```
kimfom01, line 6: If you check that link I gave you for properties you will notice that the first letter is capitalized. It's kind of a naming convention e.g. `public string FirstName { get; set;}`

I also want to point out that since you don't re-assign any of these properties outside the Habit class you can just have a get and remove the set e.g. `public string FirstName { get; }`

HabitTracker.Wolffles/Habit.cs
```diff
@@ -0,0 +1,97 @@
...
+			return result;
+		}
+	}
+	private bool CheckIfExists(string date)
+	{
+		bool commandCheck = ExecuteCommand(@$"SELECT * FROM {m_habitTable} WHERE Date = '{date}' ");
+
+		return commandCheck;
+	}
+	public string GetConnectionString()
```
kimfom01, line 40: Also we no longer need these getters it will be "Auto-Implemented"

kimfom01 (changes requested):
Everything is working fine now but I still have two more changes that I'd like to suggest 😅 before I approve the submission 

I will also like you to learn about SQL injection and how to avoid it in your .NET applications. You should try to incorporate it in your next project 😁

HabitTracker.Wolffles/Habit.cs
CWolffles: Interesting. Is it a C# specific convention? Often times in C++ I was encouraged to start all my members in a class with m_ lower case.

CWolffles: I'm also going to look into SQL Injection but most importantly wanted to say thank you for the reviews! They've been very helpful and I appreciate the work you guys do!

kimfom01: > Interesting. Is it a C# specific convention? Often times in C++ I was encouraged to start all my members in a class with m_ lower case. 

Yep it is. When you can you should check [Common C# Coding Conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions)

kimfom01 (approved):
I think it looks good overall... looking forward to your next project 😁

PR #32, opened 2023-07-03
Finished project.

alvaromosconi:


HabitTracker.alvaromosconi/HabitTracker.alvaromosconi/HabitRepository.cs
```diff
@@ -0,0 +1,405 @@
...
+            {
+                case "0":
+                    break;
+                case "1":
+                    Console.Clear();
+                    Insert(null);
+                    break;
+                case "2":
+                    Console.Clear();
+                    PrintAllExistingHabitNames(existingHabitNames);
```
TheCSharpAcademy, line 148: Could you please fix this issue? 😊 Also this is optional but I recommend you create a method when a case in the switch statement gets too bloated. You want your switch statements to be lean and easy to read. 

image

TheCSharpAcademy (changes requested):
@alvaromosconi Your project is not compiling due to the error I pointed out in the comment.

HabitTracker.alvaromosconi/HabitTracker.alvaromosconi/HabitRepository.cs
TheCSharpAcademy: @alvaromosconi Let us know if you need help 😁

alvaromosconi: Thanks for offering help, I think the app is now fine. Sorry for not having responded. I've been busy with other things these days :/

I am also making the Flashcards app. But I think I'm going to have to do this one again because when I was about to incorporate the study session part I realized that I might not make the best design decisions. Would you mind taking a look at it? it would help me a lot. 

This is the repo: https://github.com/alvaromosconi/CodeReviews.Console.Flashcards

TheCSharpAcademy:
@alvaromosconi I can't see your project in the dashboard anymore. Could you please resubmit it?

PR #33, opened 2023-07-04
just only one requirement left

Mo3ses:
made everything in 3-4 hours of code and learning sqlite, just need to finish the readme file


TheCSharpAcademy:
@Mo3ses Hi there! The project is looking good, I love the menu. I have a question though. What's the expected behavior when users choose option 1: Create Habit? I'm currently being taken back to the main menu when I choose it: 
image

Mo3ses:
so option 1 is used to track +1 record in the habit if you don't have a habit I think it returns to the menu notifying you that there is no habit and you need to create one to be tracked.
image

it may be that my version is different because I added one more type of report and a rule for validating a habit in the database I will be uploading it at the end of today along with a readme to close the project

TheCSharpAcademy (changes requested):
@Mo3ses I get it know. The project leaves it a bit open for interpretation, for sure. The idea I had in mind was something like this. Once you choose "Track Habit", you'd be presented with a few prompts. Example: "Please add the date (of when the habit was performed)", then on the next one "Please add quantity (i.e. number of cups of water, or number of pages in a book)."

If you don't let the user add the date , they can't record a habit that was done in the previous day or before that. Does that make sense? 😊

TheCSharpAcademy:
@Mo3ses Let me know if you need help!

TheCSharpAcademy (changes requested):
@Mo3ses I don't want to be a pain in the ass, but I think it would be better to remove the 3rd prompt altogether. You only need a date and a value for that date. That would be an instance of that habit. Example: 2023-07-07: 9 pages, 2023-07-08: 10 pages. The way it is right now it's a bit confusing. Let me know if you have questions. 

image

And please don't feel discouraged. At work I've had my code returned to me dozens of times before being approved. In fact even in my current job, with a few years of experience, I rarely create PRs that get approved without any modifications. This perseverance will be needed when you get your first job.

Mo3ses:
No problem, I'm here to learn, I also believe that it's very difficult for a program to go through the first time without changes and normally it should take longer than the time we're here, I'm waiting for feedback on the last update.

TheCSharpAcademy:
@Mo3ses Well done! Looking forward to see your work in the next projects 😊

PR #34, opened 2023-07-04
Project FInished!

CarDioLogic:
Habit Tracker app finished!


TheCSharpAcademy:
@CarDioLogic Well done! You were able to demonstrate you mastered basic SQL + C# 👏🏽👏🏽👏🏽👏🏽.

In regards to the database design, you wouldn't have to create a table per habit. You'd instead have a table with all instances of habits performed. And a column with a "habit id" which would be the way to identify which habit was performed. 😊

CarDioLogic:
Thanks for the feedback! :)

PR #36, opened 2023-08-07
Draft for code review

MartinL-no:


TheCSharpAcademy:
@MartinL-no I'm approving the project, well done!

I just have one thing to suggest. It's not considered good practice to handled expected behaviors in your app with exceptions. As the name says, they're designed to handle unexpected errors. For everything else it's better to handle in code (loops, if-else, etc). Not only this approach improves readability, but Exceptions are expensive. Two nice reads on this:
https://enterprisecraftsmanship.com/posts/exceptions-for-flow-control
https://stackoverflow.com/questions/891217/how-expensive-are-exceptions-in-c

PR #37, opened 2023-08-08
Finished Project

lordWalnuts:
I finally feel i can write real applications. everything from here is just abstractions on the underlying concepts. 
Be brutal in code review.


TheCSharpAcademy (changes requested):
@lordWalnuts The project is looking good, thanks for submitting. 
It does everything it's supposed to do, but you should do validation for negative entries. 
image

You could write unit tests for the validation, it's a great way to test several scenarios with a single click.   If you want to do that, check this project out: https://thecsharpacademy.com/project/21

lordWalnuts:
I've added checks for negative entries now.

PR #38, opened 2023-08-09
Final Changes

KdRome:


TheCSharpAcademy (changes requested):
Hey there! Looks like you're asking me to input the data in a format that's not valid. Could you please modify that?

image

PR #39, opened 2023-08-11
PR

ItsSt0rm:


TheCSharpAcademy (changes requested):
@ItsSt0rm Looks like the delete function isn't working, when I type in the Id, nothing happens. 🥺

Also, this isn't a requirement, so I'll still approve the project, but this query isn't readable. When you need complicated queries (I don't mind if they come from chat GPT), make sure they're readable, otherwise they're impossible to debug. 

image

PR #41, opened 2023-08-19
Initial commit

Ramseis:


TheCSharpAcademy:
@Ramseis 
image

Well done on completing the project! For future reference, it would be good to add validation to prevent a negative quantity 😊😊

PR #44, opened 2023-08-30
submit for code review

HappyReni:


TheCSharpAcademy:
@HappyReni It would have been better to have the logs recorded as integers, as it would be easier to do operations with the values. But it wasn't a hard requirement, so I'm approving it. Well done. 😁

PR #45, opened 2023-09-02
My solution

strawfervor:


TheCSharpAcademy (changes requested):
@strawfervor The project is looking great, thanks for submitting!

We need a bit more validation, though. At the moment I can input a negative amount and an invalid date: 

image

strawfervor:
ok, I've just added validation of amount of water.
Thank you for your feedback!

PR #47, opened 2023-09-07
First upload of the project.

kakurokan:


TheCSharpAcademy (changes requested):
@kakurokan Thanks for submitting! 😊
A few items need to be addressed.
1 - The app should create the database and the tables. 
2- You don't need to ask for the id from the user. Id should auto-increment.
3 - With the DB you provided, there was no unit column, so the app was throwing an error when trying to insert a unit.

kakurokan:
Now I think that is Okay

TheCSharpAcademy (changes requested):
@kakurokan Hey there, thanks for the fixes. The db creation works 😁
I found out that the delete function isn't working. Could you please have a look? Also your message mentions fruits regardless of what I'm trying to delete 😊 
image

kakurokan:
I don't understand, the delete function is working fine for me

TheCSharpAcademy:
@kakurokan Are you sure? After you delete a habit and you select to view habits, it has disappeared?

kakurokan:
Look, for me, the Habit that you want to Delete shows on the screen before you Delete all.
image
And is deleted after.
image

kakurokan:
I don't know what's happening😕

kakurokan:
> @kakurokan Are you sure? After you delete a habit and you select to view habits, it has disappeared? In my app I get a "successfully deleted" message but it doesn't get deleted. When I choose one to delete I'm taken to this menu (which doesn't really make sense):
> 
> image
> 
> There's something weird here, as you definitely don't need nested loops: image

I used the nested loops to be able to delete more than one habit per time, without a limit of habits.

TheCSharpAcademy:
@kakurokan Ok I'll double check

TheCSharpAcademy:
@kakurokan I wasn't pressing space to choose habit, I was simply pressing enter 😆😆

kakurokan:
😆😆Lol man, I was so stressed😆😆

PR #48, opened 2023-09-08
Habit Tracker Completed

tomdonegan:


TheCSharpAcademy (changes requested):
@tomdonegan Hey! Thanks for submitting! It's looking good!

The app seems to have a bug, though. When I switch to cycling and try to update an entry, it updates the entry in the drinking water table. Could you please have a look?

TheCSharpAcademy:
@tomdonegan I'm going to approve it since you demonstrated you can do stuff with SQL, but there's another bug, if I delete a habit it should move back to an existing habit or it breaks: 
image

PR #56, opened 2023-09-19
Complete HabitLogger

rthring:


TheCSharpAcademy:
@rthring well done! Great project. The only thing I would add is a message when I try to choose an Id that doesn't exist on the database. 😁

PR #57, opened 2023-09-24
Trying Again - My Solution added

Paul-W-Saltzman:
I think I got it figured out.  Feedback Welcome.


TheCSharpAcademy:
@Paul-W-Saltzman Well done! One of the best habit trackers so far, very comprehensive.

In regards to code style, a few considerations for the future:
- Remove unused usings: 
image

- Don't leave unnecessary empty lines: 
image

- Avoid comments. It would be better to create a verbose method name instead of an abbreviated name that requires further explanation. Your code should be self-explanatory:
image

You can always create summaries though, that's good practice. But comments tend to liter the code. Only use them when absolutely necessary. 

You can find [our code conventions here](https://thecsharpacademy.com/article/58):

PR #61, opened 2023-10-03
Initial Upload

mekasu0124:


mekasu0124:
I'm not sure what happened to prevent the code from going through the check system....

TheCSharpAcademy (changes requested):
@mekasu0124 Thanks for submitting! There's something happening when I'm trying to pick an existing it: 
image

Also, I'm not sure what a "current habit" means?

mekasu0124:
> @mekasu0124 Thanks for submitting! There's something happening when I'm trying to pick an existing it: image
> 
> Also, I'm not sure what a "current habit" means?

I fixed the DeleteEntry function to correct variable naming for readability and removed the need for the second database query. I edited the DeleteEntry in the Database.cs file to use the Id's gathered from the database instead of creating id's to check against using the for-loops current index count. It should work as expected now. Thank you for reviewing!

mekasu0124:
> @mekasu0124 Thanks for submitting! There's something happening when I'm trying to pick an existing it: image
> 
> Also, I'm not sure what a "current habit" means?

A "current habit" is a habit that is currently being tracked. I can change it to "Existing" instead of "current", but they mean the same thing virtually.

TheCSharpAcademy:
Ok you've met the requirements. So I'm approving your project, well done. I would have designed it slightly differently though. Ideally you want to have one or more habits. And for each of them you'd have a list of entries. Example: For the Reading habit I would have an entry on the 21st of March of 2023, 60 pages read. Then another one: 23st of March of 2023, 30 pages read, and so on. That would provide a more detailed tracker. 

But well done, nonetheless 😊😊

PR #62, opened 2023-10-03
App done.

Nik00NN:


TheCSharpAcademy:
@Nik00NN  Well done! I'm approving your project. 😊😊. For future reference, I think it would be a good idea to let users input the date. Otherwise the ability to log becomes limited, as they can't input records in case they miss a day. Please make sure you add that possibility to the next project (coding tracker) 😁

PR #63, opened 2023-10-05
Add complete project

dnym:
Full documentation at [my personal repository](https://github.com/dnym/tcsa03).


TheCSharpAcademy:
@dnym Well done! I like how you stored your queries in fields. 😁
One thing I would have added: Upon inputting an invalid date it would be good to warn the user. Something like: "Invalid date: Try again". It would also be nice to let the user know what date format was expected.

PR #73, opened 2023-10-20
Initial pull (Completed project)

cmedina-dev:


TheCSharpAcademy (changes requested):
@cmedina-dev hey there! Thanks for submitting. The project is looking good, you're showing good command of just a couple of considerations.

1 - You definitely shouldn't create a table for each habit. The way this can be achieved without a massive load on the database is having a "Record" table, which will have a HabitId property. This way you can track records from multiple habits.

2 - At the moment there's a fundamental issue with the design of the business logic. We're not really "tracking" habits if I can't visualize them across time. Ideally you'd like to know how much of a habit you've done per day. So you'd have a single table of records that would contain a quantity and a date and potentially a habitId(if you're tracking many habits).

Let me know if you have questions. 😊

cmedina-dev:
Hi @TheCSharpAcademy, issues 1 and 2 should be fixed now! Each entry is now a record within a single "Habits" table. You're able to now view all entries and their associated dates for whichever habit the user specifies.

TheCSharpAcademy:
@cmedina-dev Hey there! Well done! It's still slightly different than I imagined but you showed good command of SQL/ADO.NET and you're not creating multiple tables, so all good. 😁

For the coding tracker, users should be able to visualize their record per date. There should be a list of days with the amount of coding done each day.

cmedina-dev:
Thanks @TheCSharpAcademy! I'll be sure to take your feedback into consideration for the Coding Tracker project :)

PR #79, opened 2023-11-02
final build with simply report query

samggannon:


TheCSharpAcademy (changes requested):
@samggannon Thanks for submitting!

When trying to insert a record I get this: 
image

TheCSharpAcademy (changes requested):
@samggannon Thanks for re-submitting! 😊
It looks like you're not validating the date input. I was able to input 14 as a month:

image

PR #80, opened 2023-11-13
finish the Habit Tracker

czerviik:


TheCSharpAcademy (changes requested):
@czerviik Hey there! The project is looking good, but I found a bug by accident. 😆 When updating, if I choose an Id that doesn't exist, when I proceed to update an existing record I get this. Looks like you've closed the connection.

image

czerviik:
> @czerviik Hey there! The project is looking good, but I found a bug by accident. 😆 When updating, if I choose an Id that doesn't exist, when I proceed to update an existing record I get this. Looks like you've closed the connection.
> 
> image

Thanks for the note. Anyway I really appreciate your active supervision! If I had more friends who wanted to learn programming, I would definitely recommend the Csharp Ac.

TheCSharpAcademy (changes requested):
@czerviik Thanks for the changes. 🤓 Deleting still doesn't work though: 
image

And this isn't a requirement but it looks like we have an infinite loop here in case I only have one record: 
image

czerviik:
> @czerviik Thanks for the changes. 🤓 Deleting still doesn't work though: image
> 
> And this isn't a requirement but it looks like we have an infinite loop here in case I only have one record: image

oh it should be fixed now! I have to learn how to test the code properly. I've also found another bug when entering numbers as new habbits names and units. It was making a mess in database commands so I added squared brackets in some of those to keep the values as they should be. Maybe I could also handle those number inputs in creating new habbits.

TheCSharpAcademy:
@czerviik All good now! Well done. 🤓

PR #81, opened 2023-11-15
Initial commit

mariusgrHiof:


mariusgrHiof:
@TheCSharpAcademy Thanks for the review! The date update is to see when the log was last updated, in case a user entered wrong data. I've added a validation function so the user can't enter a date prior to the current "Date Updated" value in the database.

PR #91, opened 2023-12-30
finished project (minimal requirements)

martinka-h:


TheCSharpAcademy:
@martinka-h Well done! I suggest you try the challenges as well. They'll push the boundaries of your knowledge.

PR #92, opened 2023-12-31
Ibrahim Gilani Habit tracker console application

ibigilani:
please be ruthless with your review, i didn't implement users being able to create their own habits and I know my reports is a little basic but I wasnt sure what else to put in the reports


TheCSharpAcademy:
@ibigilani Well done! Project approved! I really like the prayer idea with a range from 0 to 5. 😎😎. I also like that you made the dates unique. Smart! And on top of that, you're using parameterized queries, which is never a bad idea 🤓

If I could add anything, is that you don't need to close the connection when using "using" statements.[ It handles that for you](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using). 
image

You could add reports such a table of prayer amounts per month, week and year. It would require you to pivot a table, which is a tricky but very nice challenge, if you're keen to push yourself. 😲😲

PR #94, opened 2024-01-17
submitting for review

frockett:


TheCSharpAcademy (changes requested):
@frockett Apologies for the delay.

When I add a new habit I'm getting this error. Do you know what could be causing this? 
image

It's good to add error handling to db calls so your app never crashes 😁

TheCSharpAcademy:
@frockett I think somehow the new habit isn't being associated with an unit_id. In terms of database design, you didn't need a table for each habit. You could have one Records table, linking to a Habits table, which will contain the unit. This way you could have infinite habits. 🤓

image


But as you said, since this is part of the challenges,  I'll approve it. 🥳

frockett:
Yeah, you're right. Thanks for pointing it out, I didn't realize that when looking back on it. I think my SQL efforts after this have been better, hopefully! Thanks, I don't mean to be lazy, I just want to take what I can from the mistakes here and keep moving since I'm a couple projects ahead now.

PR #98, opened 2024-02-04
My first projecy

LorenzoVu:
Can u let me know if is it all okay?


TheCSharpAcademy (changes requested):
@LorenzoVu Hi there! I can't see this project in our platform, could you please submit it? Also it doesn't seem like your files where committed correctly. Try to commit them and if you get stuck just created a new PR, i've cleaned up the repository.
image

PR #99, opened 2024-02-04
Habit Tracker project completed

nwdorian:


TheCSharpAcademy:
@nwdorian Well done! I think it's the first time I see anyone using Faker in this project 🤓

Just one quick comment. You don't need to close the connection when creating it with an using statement. It handles that for you.
image

PR #101, opened 2024-02-05
First pass at run habit logger CRUD console app interacting with sqlite

StanimalTheMan:
Maybe i should add screenshots to show how app works in README


TheCSharpAcademy:
@StanimalTheMan Well done! Don't worry too much, I'd usually suggest you completed the challenges but I'd prefer you get up to speed in ASP.NET Core so you can get that job 🤓🤓

PR #102, opened 2024-02-05
First Submit

Buutyful:
State pattern navigation, Sqlite DataAccess. TODO: UpdateState, DeleteState, create readme.dm


TheCSharpAcademy (changes requested):
@Buutyful I couldn't get past the menu, could you please let me know what I did wrong? 😟😟

image

Buutyful:
Hello :) You are already in the menu state and selecting a menu. I guess a more specific error message would be useful. I'll change that message to: "You are already in the menu." when menu selected in menu window and pass down the argument to the invalid command

Buutyful:
image
Now the navigation should be more clear..

TheCSharpAcademy:
@Buutyful Well done! ✅🤓

Just a couple of things: 
1 - Ideally you'd let the users update the date/time. If I miss a day using the app I can't input previous records.
2 - When using an 'using statement' you don't need to close the connection 😁
image

Buutyful:
thx for the feed back 👍

PR #103, opened 2024-02-07
Project finished

Mateusz-Platek:
Project completed


TheCSharpAcademy:
@Mateusz-Platek Well done! Project approved! For the future try to clear the console after each operation. Also it would be good to put inputs in a validation loop so you don't have to go back to the menu in case the validation fails:
image

PR #108, opened 2024-02-23
Completed Habit tracker project

Rxxyxd:
Kinda messed around with a few things on this one


TheCSharpAcademy (changes requested):
@Rxxyxd Thanks for submitting! You showed good understanding of basic SQL commands. 🤓. A few pointers:

image
This means that you'll track one habit, which means you'll need to add dates to each log. Example: If my habit is reading, a reading record would be: Feb-25, 40 pages, Mar-01, 10 pages, and so on. Otherwise it's not really a logger.

In terms of the code itself, there's a major flaw in the design. You shouldn't open a connection for the lifetime of the app and use it for every operation. If there's an error the database will be locked for a long time, crashing the app. That's not even to mention if there's concurrent use of the db. To fix that make sure you open one connection per request. You can see an example here:

image

Usually I'm not too strict with these initial reviews but the connection issue is pretty significant, so I'd rather you didn't catch a bad habit from the beginning. 🤓

Rxxyxd:
Is this sufficient for the changes requested?

TheCSharpAcademy:
@Rxxyxd That's better. I'll approve because you fixed the connection issue. 😁😁 In terms of the UX it would be better to give the user the chance to input the date as well, otherwise they miss the log if they forgot to use the app for a day. Also when deleting/updating, it's good to show a list of records 😊

PR #110, opened 2024-02-26
JanL submission

challenger720:


TheCSharpAcademy:
@challenger720 Well done! Very tidy code 🤓

One thing I should mention is that when you use 'using statements' you don' need to close the connection, that's the main job of 'using'
image

Also, for the future make sure you get rid of unused code 😆:
image

challenger720:
Thank you for your advice!😊

PR #115, opened 2024-03-04
My solution

Theosxenos:
Thanks for reviewing.


TheCSharpAcademy:
@Theosxenos Hello! I'm approving the project since you showed good command of ADO.NET 😁

For future reference, please complete ALL REQUIREMENTS, as we get more strict 👮🏻‍♂️. There was no option to delete a log. 

Also, this wasn't a requirement, but it's possible to insert a negative quantity. It would be good to add validation for that. 🙂
image

And last but not least, when asking the user to input a record reference (e.g. What date would you like to update), it's good to present a list of current records. 🤓

PR #118, opened 2024-03-15
My finished Habit Logger console app

HopelessCoding:


stevietv:
Hi @HopelessCoding - thanks so much for your submission. I've taken a look and I'm going to approve it, nice job.
I do have some tips for you going forward regarding the interface:

* consider clearing the console more often so that only the current menu shows.
* consider showing the list of records once a user choses to update or delete, so they have easy access to the id numbers
* going forward, please switch to using a file-scoped namespace, as per item 6 on our [code conventions ](https://thecsharpacademy.com/article/58)
* there is one SMALL error that currently isn't handled, if a user enters a large amount of days for the AVG calcuilation (eg 99999) then you get a DBnull error, a small sanity check there for days entered would be good :)

looking forward to seeing more of your work!!

PR #121, opened 2024-03-18
Submission HabitLogger

antonderegt:
Hi Reviewer,

Thanks for taking the time to look at my code!

PS. Codacy keeps complaining about the README.md, but it looks fine to me. I heard Codacy can have trouble with READMEs...

### DIFF
--- a/HabbitLogger.antonderegt/HabbitLogger/Program.cs
+++ b/HabbitLogger.antonderegt/HabbitLogger/Program.cs
@@ -1,0 +1,254 @@
+﻿using Microsoft.Data.Sqlite;
+
+namespace HabbitLogger;
+
+class Program
+{
+    private readonly string connectionString = @"Data Source=habitLogger.db";
+
+    static void Main(string[] args)
+    {
+        bool keepRunning = true;
+        Program program = new();
+        program.EnsureTableExists();
+        program.SeedData();
+
+        while (keepRunning)
+        {
+            ShowMenu();
+            keepRunning = program.ProcessInput();
+        }
+    }
+
+    void EnsureTableExists()
+    {
+        using var connection = new SqliteConnection(connectionString);
+        connection.Open();
+
+        var command = connection.CreateCommand();
+        command.CommandText =
+        @"
+            CREATE TABLE IF NOT EXISTS habit (
+                id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+                name TEXT NOT NULL,
+                quantity INT NOT NULL,
+                unitOfMeassure TEXT NOT NULL
+            );
+        ";
+        command.ExecuteNonQuery();
+    }
+
+    void SeedData()
+    {
+        using var connection = new SqliteConnection(connectionString);
+        connection.Open();
+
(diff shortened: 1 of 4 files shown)


HabbitLogger.antonderegt/HabbitLogger/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    void InsertRecord()
+    {
+        Console.Clear();
+        using var connection = new SqliteConnection(connectionString);
+        connection.Open();
+        SqliteCommand command = connection.CreateCommand();
+
+        Console.Write("New habit: ");
```
Doc415, line 148: Hey there i think you forgat to ask the quantity for the record to user. 
It is hard coded as 0 for all records.

HabbitLogger.antonderegt/HabbitLogger/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+namespace HabbitLogger;
+
+class Program
+{
+    private readonly string connectionString = @"Data Source=habitLogger.db";
+
+    static void Main(string[] args)
+    {
+        bool keepRunning = true;
+        Program program = new();
```
Doc415, line 12: Did you initialized the class inside itself on purpose?
You can directly call methods and fields if you make them static.

```
class Program
{
    static private readonly string connectionString = @"Data Source=habitLogger.db";

    static void Main(string[] args)
    {
        bool keepRunning = true;
        
       EnsureTableExists();
       SeedData();

        while (keepRunning)
        {
            ShowMenu();
            keepRunning = ProcessInput();
        }
    }

    static void EnsureTableExists()
    {
        using ......
```

Doc415 (changes requested):

Doc415:
> Hi Reviewer,
> 
> Thanks for taking the time to look at my code!
> 
> PS. Codacy keeps complaining about the README.md, but it looks fine to me. I heard Codacy can have trouble with READMEs...

Codacy is like a wife. It always find something to complain. You can just smile for the Readme.md ones.

HabbitLogger.antonderegt/HabbitLogger/Program.cs
antonderegt: Thanks, now you mention it, it does make more sense to keep everything static and pass the connectionString around. Also makes the methods more flexible.

antonderegt:
@Doc415 thanks for the review!

Doc415 (approved):
Thanks for your collaboration. Waiting for your next project.

Doc415:
Thanks for your collaboration. Waiting for your next project.

HabbitLogger.antonderegt/HabbitLogger/Program.cs
TheCSharpAcademy: @antonderegt Although it's ok for these beginner projects I'll point out the overusing static isn't good practice as it defeats the purpose of object oriented programming. Essentially it should only be used with classes that don't contain state. Not to mention concurrency issues. 😊

PR #122, opened 2024-03-23
HabitLogger.JaegerByte

JaegerByte:
My version of the HabitTracker. Any tips or feedback would be greatly appreciated.
Thank you for the opportunity!

### DIFF
--- a/HabitLogger.JaegerByte/Program.cs
+++ b/HabitLogger.JaegerByte/Program.cs
@@ -118,10 +118,10 @@ namespace HabitLogger.JaegerByte
             Console.WriteLine("Delete log:");
             Console.WriteLine("Please insert Id and confirm with ENTER");
             int inputIndex;
-            bool indexInput = Int32.TryParse(Console.ReadLine(), out inputIndex);
+            bool indexInputParsed = Int32.TryParse(Console.ReadLine(), out inputIndex);
             bool indexExists = entries.Any(item => item.Id == inputIndex);
 
-            if (indexInput&&indexExists)
+            if (indexInputParsed && CheckIndexExists(inputIndex))
             {
                 CommandDeleteLog(inputIndex);
                 Console.WriteLine("log deleted successfully!");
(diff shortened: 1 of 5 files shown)


HabitLogger.JaegerByte/Program.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        static void DeleteLog()
+        {
+            Console.WriteLine("Delete log:");
+            Console.WriteLine("Please insert Id and confirm with ENTER");
+            int inputIndex;
+            if (Int32.TryParse(Console.ReadLine(), out inputIndex))
+            {
+                CommandDeleteLog(inputIndex);
+                Console.WriteLine("log deleted successfully!");
```
frockett, line 119: You might want to make it such that this message is only printed when a log is actually deleted. You can do this by returning something from your CommandDeleteLog method or by checking that the ID exists and then printing the message. This isn't strictly necessary for the requirements, but I think it would be a significant improvement.

frockett (changes requested):
Hi @JaegerByte , I have only two major suggestions before approving. They should be super quick fixes. 

First, when the user wants to edit or delete an entry, no list of entries is printed. This makes it very difficult for the user to know which one they're editing.
image

Second, when editing an entry, the console tells the user that any integer is valid and allows the user to input new session data, even if the ID doesn't exist in the database. 

For example, here I am updating a session that doesn't exist:
image

And it hasn't changed anything, because there is no session with an ID of 8, even though the console told me there was: 
image

Also, it's a small thing, but you might want to insert a space between the time and the quantity. You could also print the date only, if you want. But that's just my suggestion, not strictly a requirement.

image

HabitLogger.JaegerByte/Program.cs
JaegerByte: Thank you for your input. I will fix this

frockett (changes requested):
Hey @JaegerByte , the changes look good. It's easier to understand which records I'm interacting with now. It's almost there, there are just a couple more things. The first one (which, my apologies, I didn't notice last time) is that there is no readme file, and this is the one project that explicitly requires it. The second one is that I'm still experiencing slightly unusual feedback when I try to update a log. For example: 
image

I am able to input all of the information and only then do I get the feedback that my choice was invalid. You could just add a check similar to the one you used in your delete method here: 
image

 You could extract that check into its own method which returns a bool telling you whether the entry exists or not. 

That's everything I can think of!

frockett (approved):
Great, thank you for addressing everything! Everything looks good, I'm happy to approve it 🎉

JaegerByte:
> Great, thank you for addressing everything! Everything looks good, I'm happy to approve it 🎉

Thank you for your efforts! Now, with a bit more experience, I would approach such a project a little differently :)

PR #126, opened 2024-04-04
Emgigas Habit Tracker

Emgigas:
Hello,

Please find my Habit Tracker Project for review.

Kind regards,
Emgigas


HabitTracker/Program.cs
```diff
@@ -0,0 +1,292 @@
...
+    var quantity =
+        GetNumberInput(
+            "Please enter the number of times you pet the dog on this date or type 0 to return to the menu: ");
+
+    using var connection = new SqliteConnection(connectionString);
+    connection.Open();
+    var tableCmd = connection.CreateCommand();
+
+    tableCmd.CommandText =
+        $"INSERT INTO pet_the_dog (Date, Quantity) VALUES ('{date}', {quantity})";
```
antonderegt, line 149: Try to look at parameterised SQL commands to prevent [SQL Injection](https://www.w3schools.com/sql/sql_injection.asp).

HabitTracker/Program.cs
```diff
@@ -0,0 +1,292 @@
...
+
+    tableCmd.ExecuteNonQuery();
+    connection.Close();
```
antonderegt, line 152: The using statement will take care of [closing the connection](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.close?view=dotnet-plat-ext-8.0#examples).

HabitTracker/Program.cs
```diff
@@ -0,0 +1,292 @@
...
+            $"INSERT INTO pet_the_dog (Date, Quantity) VALUES ('{GenerateSeedDate()}', {quantity.Next(0, 1000)})";
+
+        tableCmd.ExecuteNonQuery();
+        connection.Close();
+    }
+}
+
+string GenerateSeedDate()
+{
+    Random random = new Random();
```
antonderegt, line 279: Try to only create [one instance](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-random#avoid-multiple-instantiations) of the `Random` class to make the output more random.

antonderegt (changes requested):
Hey @Emgigas nice to read your lessons learned. And about not feeling comfortable with OOP and creating multiple classes, don’t worry. In the next projects you will be nudged in the right direction.

I was about to approve you project but I found a little bug. When I try to insert a record, the prompt tells me I can return to the main menu if I press 0. Cool that works. But if I then try to do something else, for example closing the application, it's not possible.  See the following output:

```
MAIN MENU

What would you like to do?

Type 0 to Close the Application.
Type 1 to View All Records.
Type 2 to Insert a Record.
Type 3 to Update a Record.
Type 4 to Delete a Record.
-----------------------------------

Option: 0
This application is now closing. Goodbye!
Invalid input format.

Please enter a date for the record (format: yy-mm-dd). Type 0 to return to the main menu: 
```

If you fix this bug, I'll approve this project because you implemented all requirements and your code is good!

Also, check for users entering negative numbers, users will (unknowingly) do everything they can to break you application 😉 . But I'm okay if you only fix the bug I mentioned above.

The other comments I added to your code are just suggestions you can play around with if you like.

I'm looking forward to seeing how you'll fix the bug!

HabitTracker/Program.cs
Emgigas: Hey @antonderegt, I think I've achieved this by removing the randoms from the seed methods and just initialising the Random class in the Program class. Hope that is what you meant.

HabitTracker/Program.cs
Emgigas: Interesting read and noted for my coding tracker project.

HabitTracker/Program.cs
antonderegt: Yes, that sounds like a good fix!

Emgigas:
Hi @antonderegt, 

That was a challenge to fix! Noticed it wasn't just that you couldn't enter any information when you returned to the menu but that the function you appeared to leave was still going even when you were at the main menu.

Originally tried to fix this by putting in a return statement at the point the the user requested to go to the menu but that didn't work as the parent function (sorry if my wording is nonsense) would still continue.

Eventually did a bit of googling and found out that I could create a try catch block which would skip the rest of the parent function at the point the 0 was entered.

Initially this was really frustrating me but really happy with the solution found.

Hope it holds together now!

antonderegt:
@Emgigas well done on fixing the bug! Getting frustrated, pulling through and getting the solution is what makes programming so awesome 😄 

APPROVED!

PR #130, opened 2024-04-13
A 'Habit Tracker' Console Application.

zen-pb:
I started implementing the Main method instead of using top-level statements, which I found really confusing and was stuck on for days. However, I broke through it and finished the project. I didn't follow the requirements by sticking to just one habit; I allowed for different habits to be inserted. This resulted in a problem where you can't insert a habit with the same name, even if it's not on the same date.

I will revisit this in the future to fix it or revamp how my code works. I just need to move forward or my brain will hate me haha.

### DIFF
--- a/DatabaseHandler.cs
+++ b/DatabaseHandler.cs
@@ -1,0 +1,141 @@
+using Microsoft.Data.Sqlite;
+
+namespace HabitTracker
+{
+    public class DatabaseHandler
+    {
+        readonly private string dbSource;
+
+        public DatabaseHandler()
+        {
+            this.dbSource = @"Data Source=habit_tracker.db";
+            this.HabitTable();
+        }
+
+        void HabitTable()
+        {
+            using var connection = new SqliteConnection(this.dbSource);
+            using var command = connection.CreateCommand();
+
+            connection.Open();
+            command.CommandText =
+                $@"CREATE TABLE IF NOT EXISTS habits(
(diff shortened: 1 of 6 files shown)


TheCSharpAcademy:
@zen-pb Have you submitted this project in the academy? I can't see it in the dashboard.

zen-pb:
> @zen-pb Have you submitted this project in the academy? I can't see it in the dashboard.

Sorry for the late reply, was busy doing other stuff. 

No, I haven't.

TheCSharpAcademy (changes requested):
@zen-pb Hi there! Thanks for submitting. 🤓👏🏻✅

At the moment there's no quantity prompt for the record. Could you please fix that? 

image

Also, you're right, it defeats the purpose of the app if I can't add multiple instances of the same habit. You need to check if there's a habit with same date and name or create two tables, one for habits and one for habit records.

And for the future, when you get stuck, it's better to come to Discord and ask for help. The projects will only get harder and you don't need to be stuck for days. 😎

zen-pb:
Hi! 

I already updated the code and added the prompt. 

Thanks for bring it up to me!

TheCSharpAcademy:
@zen-pb Hey there! Well done! I'm approving the project 🤓

For the future: 

➡️ It's better if you let the users insert the dates they want, otherwise they can't keep track of whatever they're tracking if they miss a day.

➡️ Updating should allow the user to update to any value, not just increment.
image

PR #136, opened 2024-05-16
Habit Logger Console App from Nelson

N-Endy:
This was a delightful project. I look forward to the review and corrections on how to make my code better.


VocalNight (approved):
Everything looks good!
Nothing stands out to me as wrong, the only thing that i would suggest is finding a way to clear the console after some commands, it get's cluttered otherwise.

image

Aside form that, good job! Good luck on the next projects.

N-Endy:
Thank you!

PR #138, opened 2024-05-27
Finished project.

Patryk-MM:


BBualdo:
Hello @Patryk-MM 

I like this one, good job! A lot of good practices out there. I like you took care of empty list case and documented Database methods with <summary> tags!

Only 2 tips for you:

- It's not a good practice to pass nested methods as parameters to other methods: ```db.ViewHabits(db.GetHabits(db.ChooseTable()));```.
Generally it works fine, but is less readable than storing returned results in variables and then pass them to methods. 😁

- You don't have to close connection in ```using() {}``` block. It is disposed automatically 😎

- I would add ability to delete habits as well, but that was not a requirement, so...

Approved! 🎉 Good luck bro!

PR #139, opened 2024-05-30
Habit Tracker Complete (2/3 Challenges)

Johay90:
The project is finally complete!

I have implemented most of the required functionality for the Habit Tracker application:

1. The report functionality challenge was not done, but I did attempt the other two challenges.
2. The second challenge (seeding the database) is implemented and can be triggered via the menu rather than automatically on database creation. This approach should have a similar effect.

I made a big effort to keep my code clean in this project and tried to split responsibilities into their own classes. However, there is room for improvement. For instance, the User/Console Interaction could be further split into validation and interaction components. I also did my best to follow DRY throughout the code.

Overall, this was a fun project!

### DIFF
--- a/HabitTracker.Johay90/HabitTracker/Properties/launchSettings.json
+++ b/HabitTracker.Johay90/HabitTracker/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "HabitTracker": {
       "commandName": "Project",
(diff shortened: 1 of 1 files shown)


BBualdo:
Hello @Johay90.

I like your project, because it's different. I like your idea to seed database on request, but you didn't handle issue when user wants to delete or update record with empty database. But it's a details, you just had to add ability to return to main menu or display "Press any key to return to Main Menu" when list is empty. 😁

However, you got 10 points for fresh approach to project 😅

Approving 🎉🏆🙌

PR #143, opened 2024-06-18
Project solution

tonyissa:
Had fun with this one, although I could've cleaned up the code a bit. I don't know how to use modules/assemblies yet in C# so that would help as well as breaking down some of the more repetitive code into functions.


stevietv:
hi @tonyissa - thanks for your submission. Right now it needs a little bit of work before it can be approved:

* I can cause the app to crash by entering the Delete or Update menu and entering an empty id (i.e. just pressing enter). One of the project requirements is that the app doesn't crash, so you should be handling these errors.
* make sure to submit your code in a valid foldername within the repo, eg `HabitTracker.Tonyissa` - so that the files dont appear in the main folder.

tonyissa:
> hi @tonyissa - thanks for your submission. Right now it needs a little bit of work before it can be approved:
> 
> * I can cause the app to crash by entering the Delete or Update menu and entering an empty id (i.e. just pressing enter). One of the project requirements is that the app doesn't crash, so you should be handling these errors.
> * make sure to submit your code in a valid foldername within the repo, eg `HabitTracker.Tonyissa` - so that the files dont appear in the main folder.

Thanks for the feedback. I've implemented some basic error handling so the app doesn't crash, as well as fixed that error.

tonyissa:
Really not sure what's going on with the readme file and why codacy doesn't like it, but I'll have to check it out later.

stevietv:
thanks for the update, this is now approved! ✅ 

You can in future ignore codacy warnings for ReadMe files

PR #144, opened 2024-06-19
Finished project.

jollejonas:
Looking forward to reading the review, thanks!

### DIFF
--- a/jollejonas.HabitTracker/README.md
+++ b/jollejonas.HabitTracker/README.md
@@ -1,0 +1,40 @@
+# ConsoleHabitTracker
+C# Console Application to track habits.
+It uses SQLite database to store the data.
+The functionalities include:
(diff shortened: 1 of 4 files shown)


stevietv:
hi @jollejonas - thanks for your submission and great work on attempting the challenges. The basic requirements have all been met so this is approved ✅.
There are a few small issues with your challenges that I'll give you feedback on:

* you should never let a user create tables in your database, as you are doing with the new habits created. This can lead to massive database growth and potential hacks due to vunerabilities. You should have tackled this here by having a table "habit_types" where the user can add a habit name and unit of measure, and linked your tracked habits table to that. This would result in a relational database and thus is a pretty complex challenge.
* even when choosing a different habit I'm asked how much water I drank, which doesn't make sense!

PR #145, opened 2024-06-20
Habit tracker console app

ananttuli:
Implemented the habit tracker console app. (Haven't attempted the challenges). 
For some reason I'm getting 1 lint issue only for the `README.md` file on codacy, hope that's okay.

- [x] This is an application where you’ll register one habit.
This habit can't be tracked by time (ex. hours of sleep), only by quantity (ex. number of water glasses a day)
- [x] The application should store and retrieve data from a real database
- [x] When the application starts, it should create a sqlite database, if one isn’t present.
- [x] It should also create a table in the database, where the habit will be logged.
- [x] The app should show the user a menu of options.
- [x] The users should be able to insert, delete, update and view their logged habit.
- [x] You should handle all possible errors so that the application never crashes.
- [x] The application should only be terminated when the user inserts 0.
- [x] You can only interact with the database using raw SQL. You can’t use mappers such as Entity Framework.
- [x] Your project needs to contain a Read Me file where you'll explain how your app works.

### DIFF
--- a/Console.HabitTracker.ananttuli/Console.HabitTracker.ananttuli.sln
+++ b/Console.HabitTracker.ananttuli/Console.HabitTracker.ananttuli.sln
@@ -1,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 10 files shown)


stevietv:
Hi @ananttuli - thanks for your submission, great code and a beautiful readme! this is approved, good work ✅ 

I do have a couple of suggestions for you for consideration later:

* Asking a user to input a date in format `2024-01-01` is a bit ambiguous is that yyyy-mm-dd or yyyy-dd-mm - consider this in future
* the report presents all records at a time of 00:00 - consider trimming this from the results
* consider showing the user the list of entries when you choose to edit or delete, otherwise I need to know an ID!

ananttuli:
@stevietv Thanks for the review & feedback!

Great suggestions - Implemented these in my fork and for my CodingTracker PR submission https://github.com/TheCSharpAcademy/CodeReviews.Console.CodingTracker/pull/102

PR #150, opened 2024-06-21
Finished project with 3/3 challenges

jrazorx:


stevietv:
@jrazorx - don't forget to submit this in the Academy Dashboard before it can get reviewed

stevietv:
hi @jrazorx - what a fantastic submission! great work in meeting all the requirements AND all the challenges. Loving the clean and structured code that is also easy to follow. and that ReadMe👌🏻!!!

There is one issue that I'm not going to decline you for as its part of the challenge - currently deleting a habit type fails when there are entries for that habit, as you havent set the `HabitTypeId` field to `ON CASCADE DELETE`.

Also, a tip for future - give the user the option to also enter the date, what if I wanted to log yesterday as I forgot?

I look forward to seeing more of your projects!!

PR #153, opened 2024-06-23
Project finished!

pinksparrows:


Habit-Tracker.pinksparrows/Habit-Tracker/Program.cs
```diff
@@ -0,0 +1,366 @@
...
+        Clock();
+        Console.WriteLine("------------------------------------\n");
+        Console.WriteLine("Insert -> 1\n");
+        Console.WriteLine("Delete -> 2\n");
+        Console.WriteLine("Update -> 3\n");
+        Console.WriteLine("View -> 4\n");
+        Console.WriteLine("Exit -> 0\n");
+        Console.WriteLine("------------------------------------");
+        Console.WriteLine("Please provide your input below:");
+        getInput();
```
stevietv, line 37:
```suggestion
        GetInput();
```

stevietv (changes requested):
hi @pinksparrows - I'm gonna review your project, but first please fix this part so that the project runs! Then I will take a look at it :)

pinksparrows:
> hi @pinksparrows - I'm gonna review your project, but first please fix this part so that the project runs! Then I will take a look at it :)

Sorry about that! I thought I had changed that already, should be alright now :)

stevietv:
Hi @pinksparrows - I've reviewed it and whilst its a really good start at tackling databases in .net, right now it doesn't really capture the requirements of the project entirely. I think there may be some misunderstanding on what is required.

Firstly, You should never let users be creating tables in your database. This is very risky.
You should be creating a table that allows the user to record instances of habits being tracked. Typically this table would contain the following columns:

| Habits |
| ----- | 
| Id |
| Date | 
| Quantity | 

If you choose to tackle the challenge of letting user also choose habits then you'd have two tables:

| Habits|HabitsLogged|
|--------|--------|
| Id | Id |
| HabitName | HabitId (links to habits table) |
| Unit | Date |
|  | Quantity |


Right now in your database I have to enter the habit name exactly each time, which is open to error, it should be possible to select by id number. Also if I have more than one row in a habit table its impossible to edit any subsequent rows.

Sorry to be harsh, but I'd like you to take a look at this one again, perhaps firstly just focus on the single habit without allowing the user to make multiple habits (as that requires a relational database).

pinksparrows:
Sorry for the long wait! I had been a little occupied with other projects for a bit. I believe I changed it correctly now, where the table is created upon launching. If there's anything I miss or did wrong I appreciate the advice!

stevietv:
> Sorry for the long wait! I had been a little occupied with other projects for a bit. I believe I changed it correctly now, where the table is created upon launching. If there's anything I miss or did wrong I appreciate the advice!

Hey @pinksparrows - I can see in your program how to create a habit, but how do I actually then log an entry against that habit? like date and how many times? I get the feeling that it might be valuable for you to watch the video on the project page to get a good idea of what the project expects.

stevietv:
Hi @pinksparrows - The update brings the app in line with the requirements, thanks for tweaking it!
Only thing missing now is the final requirement on the [project page](https://thecsharpacademy.com/project/12/habit-logger):

> Your project needs to contain a Read Me file where you'll explain how your app works.

Could you add a readme for this, and then we'll finish the approval.

pinksparrows:
Oh wow, almost forgot that! I guess it didn't come with the pull request before. I created it within my project folder so it should be in now.

stevietv (approved):

PR #159, opened 2024-07-03
Console.HabitTracker.cjc-sweatbox

cjc-sweatbox:
Implemented as per tutorial and then expanded to include the challenge requirements.


javedkhan2k2:
@cjc-sweatbox great work. A very good structure and well thought application with user friendly UI.
It is approved. I only found small bugs which you can easily fix.

Date ranges are not validated in reporting. e.g. end date should be after start date.
The inputs should be trimmed before validating.

cjc-sweatbox:
@javedkhan2k2 that you for your review and input. I will raise an issue on my actual repo and fix.

Thanks
Chris

PR #162, opened 2024-07-11
done

kwm0304:


TheCSharpAcademy:
@kwm0304 Well done! Congratulations on completing the challenges! That's how you learn faster! ✅😁

I'm going to approve since you demonstrate understanding of ADO.NET but I was expecting to see a log of a record per date. Example: For the running habit, I'd like to see a list of days with how much I've run for each day. Otherwise all I have is the final is a sum, not a log. Hope that makes sense?

I also found this instance where you're using an exception to handle an expected scenario (when there are no records). Exceptions should be used for ... exceptions. They're expensive. Here a simple message to the user would suffice.
image

Other than that the code is well organized and following "Clean" principles 👏🏻😎✅

kwm0304:
I didn't even consider that, listing the entries out would have been a better way to go. Why did you close and not merge?

PR #163, opened 2024-07-12
My Habit Tracker

echoblade100:
[My Habit Tracker README](https://github.com/echoblade100/HabitTracker/blob/main/README.md)


TheCSharpAcademy (changes requested):
@echoblade100 Well done! Your code is super organized and you've completed the challenges. Great work! 

I was about to approve but I've realized you're keeping a connection open through the lifetime of the app, which is a bad practice. I'll ask you to refactor your code to use a connection per transaction before I can approve it. 

Also a few things you can optionally fix:

It would be good to make the records visible when asking for the id. 
image

Here it would be better to let the user know the exact format. 
image

Don't forget to remove unused code:
image

echoblade100:
Hello. I remade the connection to the database, there is one connection for each request. Rewrote the input description for entering a date. I hope it became more clear. I have removed unnecessary garbage from the code.

TheCSharpAcademy (changes requested):
@echoblade100 Quick question, why are you creating readers for every operation?

A reader is typically used to retrieve data from the database, not to perform operations that modify data. 🤔🤔

echoblade100:
I return it just to make sure that the operation was completed. Or somewhere in some place I will need the newly changed object after the update. I don't know much, so I don't know if it's unnecessary to return an object or if it's just enough that the operation completed without an error.🤷

TheCSharpAcademy (changes requested):
@echoblade100 Sorry to be a pain, but It's definitely unnecessary. This violates the YAGNI principle (You're Not Going to Need It). Only return things you actually need. 😁 This would be heavily frowned upon if a prospective employer assessed your code. Could you please update before you can approve? I'm happy for you to ask chat gpt why it's a bad practice. It will come up with comprehensive explanation.

echoblade100:
@TheCSharpAcademy I removed unnecessary functionality. But I left the return in the `CreateHabit` method. Because I pass an [object without an id](https://github.com/TheCSharpAcademy/CodeReviews.Console.HabitTracker/pull/163/files#diff-2f73aff13b38a9a347ca0a429cd9539e8fc3356284b7e72cbadf03bf58b61931R27-R28) for creation and I have no other way to return it after creation and get the id provided by the database.

TheCSharpAcademy:
I'm going to approve the project because I don't want to slow you down, but I don't think you understood what I meant. Let me explain. Every time you create update or delete data you're creating a data reader and reading every record from the database: 

image

This is inefficient and absolutely not necessary. The correct would be something like this:
image

No data reader is created, and "command.ExecuteNonQuery()" is called.

Hope that makes sense? 🤓

echoblade100:
@TheCSharpAcademy  Finally I understood what you meant in the screenshot with the code. Thank you very much for the code review. I will try to follow principles DRY and YAGNI, KISS in the future, I hope my code will become cleaner.

PR #165, opened 2024-07-14
Project finished! With challenges requirements

NgocAnhDo26:


TheCSharpAcademy (changes requested):
@NgocAnhDo26 Well done! The project is looking great 😁

I found a bug though, you should validate for negatives: 
image

And most importantly. You shouldn't have a single connection for the lifetime of the app. It's bad practice and will cause many problems down the road. Please refactor your code so it creates a new connection per transaction, preferably with using statements. 😊

You [can find examples here](https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/ado-net-code-examples) or look into other people's Pull requests. Chat GPT also does a good job at explaining it. 🤓

TheCSharpAcademy:
@NgocAnhDo26 Let me know if you have questions

NgocAnhDo26:
> @NgocAnhDo26 Let me know if you have questions

Thank you so much for your suggestions! I didn't know that you should create new connection for every operations, I will keep this in mind.
I have also made changes to both the negative input bug and the connection problem in my newest commit! Please take a look.

TheCSharpAcademy:
@NgocAnhDo26 Thanks for the modifications. Project approved! ✅👏🏻😁

I found another issue that you'll need to fix for future projects. One of the SOLID principles is the "Single Responsibility". I noticed you're using the same method to insert records and record_habits. While the solution works, it's a bad practice. It makes the code less testable, readable and maintainable. For the future make sure your methods do one thing and one thing only. 😁

Looking forward to your next projects!

PR #167, opened 2024-07-17
Habit Tracker with all challenge done!

JerryCJY:
To whoever reviewing the code, I am sorry that you have to read the mess I made and the library I use. I definitely over-engineered this. But using the Gui library was a lot of fun. The next project will be a lot simpler.


TheCSharpAcademy:
@JerryCJY Hey! Well done! The project looks great. You solved it in a very unique way. 😁😎.

In terms of the business case itself, it would be good to let the user adds records for each habit. For example a list of dates showing how many kms I ran each day. Food for thought for the next projects. 😁

In terms of the code itself, although everything works find, I'd say it doesn't follow "Clean" principles in your Connection service. Let me explain. 

1 - I can't quickly understand what's going on in the Init() method. What are we we initiating? Upon investigation I realise it Creates tables but it also inserts data. This method alone violates 3 Clean Code principles:

Meaningful Names: Use descriptive and clear names for variables, methods, and classes.
Single Responsibility Principle (SRP): Each class or method should have one responsibility.
Keep Methods Small: Methods should be short and do one thing.

2 - Then we go into ExecuteDbCommand. Again, there's a lot going on here. What's the Db command for? Using a switch statement for all the different commands is clever. You can definitely handle all games in one method. But then it goes into "GetAsDataTable". Get what? 😁 It would require several iterations of running the code using the debugger to fully understand what's going on. Personally I think code should read like a book that a 4-year old could understand. 🐣🐣

Overall it's a great project and I can see you have great skills. For the future I recommend you look up Clean Code to keep improving even more. Even a quick Chat GPT research will do, [but you can also read the book](https://amzn.to/3Lv7Uyt). 😁

JerryCJY:
Thank you for the review, I will take some time to read the book.

PR #169, opened 2024-07-18
Project Finished

PokingOutsideOFF:
Well the CRUD operation was done successfully. However, for the report functionality I have added only 3 queries, hope that works. Please provide suggestions how I can do it better.

### DIFF
--- a/HabitTracker.sln
+++ b/HabitTracker.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.3.32811.315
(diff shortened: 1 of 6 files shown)


TheCSharpAcademy (changes requested):
@PokingOutsideOFF Hey there! It's looking good, thanks for submitting! 🤓

Could you please add a date for each record, otherwise the users can't really track a habit across time, only random instances. Should be a quick fix, before I can approve! Let me know if you have questions 😊

TheCSharpAcademy:
@PokingOutsideOFF Well done! 🤓

For the future, give the users the chance to input the date, otherwise they will fail to keep track of whatever they're logging in case they don't log it immediately. 😁

PR #170, opened 2024-07-22
project complete

Bkohler93:


TheCSharpAcademy:
@Bkohler93 Well done! Project approved! ✅👌🏻😎 Very well organized. 

Just one thing for the feature. Usually I/O resources working with the file system are unmanaged. Stream Readers such as this one for example:

image

The good thing is that you can wrap them in a using statement as well and it will handle its disposal (using var reader .......)

Keep up the good work! 👏🏻👏🏻👏🏻

PR #173, opened 2024-07-29
Submitting habit tracker for review

gafoorraees:


ukpagrace:
Hi, Nice work.
All requirements were met, Although the string input wasn't validated for numbers.
I was able to insert a number.

I also got this error while updating
System.InvalidOperationException: 'ExecuteNonQuery can only be called when the connection is open.'

gafoorraees:
Hi,

Thanks for reviewing the code! I appreciate it. I fixed the string input validation so that numbers should no longer be allowed entry when inserting or updating. I was unable to replicate the exception, but I removed the Close() method and allowed the using 
statement to automatically end the connection. Hopefully that fixes the issue.

I will begin to separate classes with the next project.

Thank you!

TheCSharpAcademy:
@ukpagrace was this project approved? You can close with comments if so. 😁

PR #177, opened 2024-08-07
habit logger project ready to review

mohammad-premtech:


TheCSharpAcademy:
@mohammad-premtech  Well done! Project approved! 👌🏻🤓✅ And well done for completing the challenges!😄

There's one thing I would have done differently. I noticed you have your validation with a few levels of nested if statements. I suggest you avoid doing that as it becomes hard to maintain very quickly. Instead just do the validations sequentially, using a loop for each.

PR #178, opened 2024-08-08
Habit Tracker for review

Joe-Whitehead:
With additional challenges (except AI)


TheCSharpAcademy:
@Joe-Whitehead Well done! Project approved! 👌🏻🤓✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster 😄

A couple of thing I'd suggest:
➡️For the creation of multiple habits you'd have two tables: Habits and Records. The approach you chose is inefficient because you don't want to let the users create tables. I suggest you have a look at **_One-To-Many relationships_** in SQL to understand how it works.

➡️When retrieving records in any application you want to use **_pagination_**. Meaning you should only retrieve a few records and let the user choose a different "page" to see more records. As soon as you have more than a few thousands records, retrieving them every time from the DB will be very expensive.

Keep up the good work! See you in the next project 🤓

Joe-Whitehead:
Thank you for the feedback @TheCSharpAcademy I will work to implement your suggestions in the next challenge (Coding time) as that is very similar. Appreciate all the help and support so far!

PR #182, opened 2024-08-14
Habit Tracker Project Submission

thomas-henley:


N-Endy:
Nice Job!

The project works as intended. I also see you implemented the challenges too. Bravo!👍

Tips: It would be good for the user to receive some feedback from the app on tasks done, e.g. when a user adds cardio, the app should prompt the user that the action has been done.

Good luck with your next project!

PR #189, opened 2024-08-23
Create: A habit tracker app

MefDev:
Implemented Requirements:

1. Simple
- [x] When the application starts, it should create a sqlite database, if one isn’t present.
- [x] It should also create a table in the database, where the habit infromation will be stored.
- [x] You need to be able to insert, delete, update and view your logged hours. 
- [x] You should handle all possible errors so that the application never crashes 
- [x] The application should only be terminated when the user inserts q or if he does not say anything.
- [x] You can only interact with the database using raw SQL. You can’t use mappers such as Entity Framework
2. Challenges 
- [x] using parameterized queries to prevent SQL Injection.
- [x] Seed Data into the database automatically when the database gets created for the first time, generating
	a few habits and inserting a 20 records with randomly generated values.
- [x] Let users choose an option using their voice leveraging Azure Language Services.


chrisjamiecarter (changes requested):
Hey @MefDev,

Are you able to include a solution file (.sln) for this project? 
I am having trouble opening the project in Visual Studio without it...

Thanks,
@chrisjamiecarter

MefDev:
Hey @chrisjamiecarter,
The solution file has been added, could you check if the project works properly now?
Thanks,
@MefDev

chrisjamiecarter (changes requested):
Hey @MefDev,

Well done on this application submission.

Would I be correct in assuming English is not your first language?

I do have some points that I believe need addressing before I can continue my review:
- Why can you not enter a habit in the past?
- When viewing, updating, deleting a habit, is the expectation that the user has to remember the ID? Currently there is no functionality to view all habits in the database?
- What is the date format when adding a habit? I entered the ISO 8601 date format and got an error (yyyy-MM-dd)

I look forward to your response!
Thanks,
@chrisjamiecarter

MefDev:
Hey @chrisjamiecarter ,
 Yes, English is not my first language. I'm sorry if my English is not understandable. 
-  Can you not enter a habit in the past?
 
Yes, my app wasn't expecting a user to enter a habit in the past, but now I removed that, you can start a habit whenever you like.
- When viewing, updating, deleting a habit, is the expectation that the user has to remember the ID? Currently there is no functionality to view all habits in the database?
 
Yes, I'm expecting the users to remember their IDs, I know that is not a good experience for the user, I tried to fix that. However, since I'm using entities I only get the entity ID instead of the ones on the database. This is why, I did not display id to the user when it has been created. In addition, previously, the user was unable to view all their habits at once. Now, they can view all their habits right away.
- What is the date format when adding a habit? I entered the ISO 8601 date format and got an error (YYYY-MM-dd)

In previous code, I used the format (MM/DD/YYYY), so I had to change it to ISO 8601 (YYYY-MM-DD)  to not have any confusion in regards to Date format.

- If you would like to check Azure Speech services you need to follow the steps on the updated readme file. [readme](https://github.com/MefDev/CodeReviews.Console.HabitTracker/blob/master/Console.HabitTracker.mefdev/README.md)

I look forward to your response!
Thanks,
@MefDev

chrisjamiecarter (approved):
Apologies, Your English is great! Very understandable.
It is just that I noticed some typos in the code and didn't want to seem too harsh. 

No worries on the points I raised! I had a look and they are not requirements, I just wanted to make sure of expected behaviour. But I like the changes you have made to the UX (user experience).

Hopefully the questions I asked have made you think and improve when designing and writing code.

Well done, I will go ahead and mark as approved, keep it up on the next projects! 😊

Best regards,
@chrisjamiecarter

MefDev:
Yes, the questions were really good, they made me think about other ways to design and write my code.
Thank you so much for raising those insightful questions. I've learned a lot.
Have a great day!

PR #202, opened 2024-09-16
Habit Tracker completed

Burnie97:
Please be honest in code reviewing! Thanks 😄

P.s I am not sure why I get one Codacy error with readme file, I've spent 4 hours trying to fix it, but guys in discord replied that its fine since its just minor style error. But anyway, I woder whats wrong with that, so if you know why it happens, please also mention it

Thank you!


TheCSharpAcademy:
@Burnie97 Don't worry about Readme errors 👌🏻👌🏻

TheCSharpAcademy:
@Burnie97 I'm going to approve the project since it fulfils all requirements, but there's a very important error in your challenges. You shouldn't generate a table for each habit. Instead you should have one habits table and one records table. A habitId foreign key on the record table will identify what habit the record is for. It's not a good idea to let the users create tables in your database. 

A smaller issue I found is that here you don't need this validation. The default case of the switch should suffice. Instead of having an integer as your switch statement you'd have a string or character, so no need for parsing.
image
Also there's a few coding style minor issues. Unnecessary empty lines, unnecessary comments and lack of separation between methods. You can [check our code conventions here](https://thecsharpacademy.com/article/58/code-conventions) 🤓
image

Other than that well done! See you in the next project

PR #219, opened 2024-11-02
Initial commit

tony-phan:


arashi256 (changes requested):
HI there! You're very close here, but unfortunately you are missing a couple of requirements. First, you need to include a readme file as per this requirement: "Your project needs to contain a Read Me file where you'll explain how your app works." Here is a link on an example readme: - https://github.com/thags/ConsoleTimeLogger . The other requirement is that while you check input for dates (nice regex, btw!) you do not for other aspects of the data entry so that it's possible to enter empty habits. Please see screenshot for example: - 
image

If you solve these two things, I will be happy to approve your project! Looking forward to see your revisions!

arashi256 (approved):
Hello again :) Good job, I am approving your submission! One thing to note going forward, please stucture your project like <project name>.<user name> rather than just <project name>. It helps keep things neat for the repos. But well done! See you in the next project 👍

PR #221, opened 2024-11-03
Completed habit tracker

yemiOdetola:
Hi there, 
Please take a look at my project when you have the time. 
Any suggestion or feedback will be greatly appreciated 🙏🏾


arashi256 (approved):
Great, works well! Some points, however: - You're automatically creating 100 seed records each time the application starts, not just on creation. So first time I ran it, I had 100 records, second time I ran it, 200 and so on. Also, the application takes quit a long time to do this, like 10-20 seconds even on my monster PC :) If you're going to have the application pause like that, you really need a message to the user on what it's doing, for a moment there, I thought it had crashed until I figured out what it was doing.

PR #222, opened 2024-11-07
habit tracker + 2 challenges

lynnthegrinch:
I wan't able to do the first challenge and the last but i managed to autopopulate the database and allow users to input any habit they want to track!


hasona23:
Hi @lynnthegrinch ,
how are you doing. congratulations on doing the Habit Logger project.😊.
1-The readme is good and well designed 
2-You didn't repeat code and applied the DRY principle which is clearly show in the input part.  


The project met all the Requirements although I believe there is a little room for enhancement 😊. 
First of all :
1- Make your code organised separate the project into different units to make it more readable and easier to debug for instance put the Models or in this case the HabitRoutine class in a folder and the database stuff in a separate folder.
2- 
image
in code comments try to avoid abbreviations such as n(and) , pls(please) and etc to appear more professional.

Project accepted.  🎉

PR #230, opened 2024-11-26
Code Review - Habit Logger

natephor:


habitLoggerApp.Natephor/HabitManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        var command = connection.CreateCommand();
+        command.CommandText =
+            """
+            SELECT Id, Date, Body, Quantity FROM habit
+            """;
+
+        try
+        {
+            using var reader = command.ExecuteReader();
+            if (reader.HasRows)
```
chrisjamiecarter, line 129: 🟡 **Readability**

💡 Contensious one here, so marked in yellow. But, `reader.Read()` will only be true if `reader.HasRows`, so one could argue that the `reader.HasRows` only introduces a redundant layer of indentation.

habitLoggerApp.Natephor/UserInterface.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public void DisplayMenu(bool clearConsole = false)
+    {
+        if (clearConsole)
+            Console.Clear();
+
+        Console.WriteLine("Main Menu");
+        Console.WriteLine("==========================\n");
+        Console.WriteLine("What would you like to do?");
+        Console.WriteLine("1 - View All Habits");
```
chrisjamiecarter, line 51: 🟠 **Inconsistent Casing**

💡 Present a consistent UI. `1 - View all habits` vs `1 - View All Habits` then `2 - Create a new habit` vs `2 - Create a New Habit`

habitLoggerApp.Natephor/UserInterface.cs
```diff
@@ -0,0 +1,71 @@
...
+        while (string.IsNullOrEmpty(userInput))
+        {
+            Console.WriteLine(message);
+            userInput = Console.ReadLine();
+        }
+
+        return userInput;
+    }
+
+    public int GetUserNumberInput(string message)
```
chrisjamiecarter, line 17: 🟠 **Validation Improvements**

💡 Can you do negative quantities? Apparently you can 😄 !

image

habitLoggerApp.Natephor/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+                break;
+            case "2":
+                var body = userInterface.GetUserInput("Enter habit body: ");
+                var parsedDate = userInterface.GetUserDateInput("Enter habit date: ");
+                var quantity = userInterface.GetUserNumberInput("Enter quantity: ");
+
+                habitManager.CreateHabit(parsedDate, body, quantity);
+                break;
+            case "3":
+                var updatingId = userInterface.GetUserNumberInput("Enter the id to be updated: ");
```
chrisjamiecarter, line 28: 🟠 **No ID Validation**

💡 You have not validated that this ID exists. Meaning that the rest of this statement block needlessly executes.

image

habitLoggerApp.Natephor/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+                var updatedBody = userInterface.GetUserInput("Enter habit body: ");
+                var updatedDate = userInterface.GetUserDateInput("Enter habit date: ");
+                var updatedQuantity = userInterface.GetUserNumberInput("Enter quantity: ");
+                habitManager.UpdateHabit(updatingId, updatedBody, updatedDate, updatedQuantity);
+                Console.WriteLine("Habit updated successfully!");
+                break;
+            case "4":
+                var deletionId = userInterface.GetUserNumberInput("Enter habit id for deletion: ");
```
chrisjamiecarter, line 36: 🟠 **No ID Validation**

💡 You have not validated that this ID exists. I know nothing has been deleted, but it still tells me that it successfully has 🙃.

image

habits.sqlite
chrisjamiecarter, line 1: 🟠 **Empty File**

❓ Delete?

chrisjamiecarter (approved):
Hey @natephor 👋,

Excellent work on your HabitTracker project submission 🎉!

I have performed a peer review and have added some praise and suggestions as bespoke comments.
Also, see below for some general feedback.

***

🟠 **Solution Structure**
As per the academy's [Code Reviews](https://thecsharpacademy.com/article/52/code-reviews) page. You project submission should be as per the below.
```
.
└── [ProjectName].[GitHubUserName]/
    ├── [SolutionDirectory1]/
    ├── [SolutionDirectory2]/
    └── [ProjectName].sln
```
This will make it easier for code-reviewers to pull down and review your code in the future. In your case, make sure everything is inside your project submission folder. This will become more obvious why as you get to the later projects.

🟠 **Console UI**
Your console window gets very busy and cluttered. You could clear the console each time before you write the menu. This would allow the user to segment an action, maintain readability of the console window and improve the experience for the user.

🟠 **User Experience**
When updating/deleting a habit, the user has to remember the ID of what they want to perform the action on.
When updating, they have to remember the previous object values/details. This makes the database redundant, if it all needs to be in the users head anyway.
When asking a user to select an ID: Give them a list of valid options. Validate that option straight away so they don't waste their life inputting something that won't take effect.
When updating an object, display the current object to the user. Give them the option to keep the existing value, so they only update what they need to.

***

I suggest revisiting to try understand and fix any issues 🔧 , but as these do not relate to the requirements they will not block my approval.

Feel free to reach out if you have any questions or if you'd like to collaborate further on any of these points 🆘 .

I will go ahead and mark as approved, keep up the excellent work on the next projects! 😊

Best regards,
@chrisjamiecarter 👍

PR #231, opened 2024-12-01
First submission

Radicals27:


arashi256 (changes requested):
Hi. This needs a bit of work, I'm afraid! 
It takes about 10 - 20 seconds to seed the data and while this is happening, there is just a blank screen. I was about to give up until I realised what it was doing. SQLite is SLOW, so you need to tell your user what it is doing. I almost force-quit the app :) Also, this seeding happens *every time* you start the app and it adds 100 records each time. This should only happen once.  Also, when attempting to update a record, the UI is confusing about the units - 
a
One further thing to note, when attempting to update/delete a record with an ID that doesn't exist, it's not clear if anything has happened, you should put in something like "record deleted" or "record ID not found" or something to make it easier to tell what the app is doing. This is not a request, mind - just something to think about :)

Radicals27:
I'll fix these now, thanks!

arashi256 (approved):
Approved. Good job :)

PR #263, opened 2025-02-12
Initial pull request for Habit Logger

fatihskalemci:


TheCSharpAcademy:
@fatihskalemci Project approved! 😄✅ And well done for completing the challenges! Fantastic work 👌🏻

🔍️Your code is super clean and everything does what it's supposed to do. There's one thing I would have done differently. For new habits I'd create a different table, leveraging the power of relational data in SQL. So you'd have a "Habit" and a "HabitEntry" table. Each habit entry would reference a habit in the "Habit Table", based on the HabitId, which would be a foreign key. The way you solved this isn't wrong, but the suggested solution would make the project easier to expand in the future. 😎

✋🏻Keep up the great work. Looking forward to seeing your next projects!

PR #265, opened 2025-02-13
first commit of Habit Tracker

RyanW84:
Not finished

Got it working
Added the ability to log more than one kind of habit.

Not yet implemented seeding


TheCSharpAcademy (changes requested):
Hey @RyanW84 thanks for submitting! It's looking good. I'm going to request changes because update doesn't seem to work. See picture below. I tried to update the record with index 1 and it didn't pick the changes up:

image

Since you're going to review it, a couple of extra things you could do
- Clear the console after each operation. That will improve readability. 
- You're currently supporting multiple habits but it would be good if you did that using two tables: HabitRecords and Habit. They'll be linked by a foreign key and you'll know which habit the record is for by referencing the Id of the habit. 

Using two tables isn't a requirement but it's a very important technique to learn as early as possible in your journey 😎

TheCSharpAcademy:
@RyanW84 Well done sir! I think I found the bug we talked about. In your if else statements it seems like there's no 'if (!updateDate and updateHabit ........)'  option. But let's move on , as you'll keep practicing SQL anyways in the next projects. 😄

PR #268, opened 2025-02-20
Redid project

mrgee1978:
Decided to totally redo the project. I learned a lot and had a lot of fun!

### DIFF
--- a/HabitLogger.mrgee1978/HabitLogger.mrgee1978.sln
+++ b/HabitLogger.mrgee1978/HabitLogger.mrgee1978.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.13.35806.99 d17.13
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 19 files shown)


TheCSharpAcademy:
@mrgee1978 Hello! Have you submitted [in the dashboard](https://www.thecsharpacademy.com/Dashboard)? 😄

mrgee1978:
Yes I have

TheCSharpAcademy:
@mrgee1978 Well done! everything works nice and I can see you completed some of the challenges as well. Great work! 👏🏻👏🏻

Since this project had already been reviewed, you won't be getting extra points for this one 🥲

🔍The only feedback I could give is that your validation classes implementations seem to have identical methods? If that's the case, wouldn't it be better to have them in the base class? Or, do we need a base class at all? This is a clear case of **YAGNI** to me, unless you know for sure there will be variations in upcoming implementation.

See you in the next projects! 🫡

mrgee1978:
Thank you so much, and you are 100 percent correct about that. I will take that into account for the next projects, though I have already submitted another, but I will take your advice into account in the upcoming projects. I am always looking to learn and improve . Thank you so much again and I look forward to doing more and more challenging projects 😁

PR #272, opened 2025-03-04
Habit Logger project

Favii:


TheCSharpAcademy (changes requested):
@Favii Thanks for submitting. It's looking good! However updating a habit isn't working, it doesn't let me choose a new name. Check this out: 
image

To support multiple habits the best practice is to use two related tables. Something like 'habits' and 'habitRecords', linked by a **FOREIGN KEY** establishing a **ONE TO MANY** relationship.

TheCSharpAcademy:
@Favii   Project approved! 😄✅ It's a shame you didn't complete the challenges this time. 😟 We strongly recommend you try the challenges. It will help you learn better and faster!

🥋And well done! You are one project away from getting the Olive Green Belt! 🫒🫒

🔍To your question, the HABIT table would have the name of the habit and the unit, along with an Id, the HABITRECORD table would have the date, the amount and the habitId, which would link to the HABIT table. That's the best way to enforce data integrity 🫡

PR #274, opened 2025-03-06
Habit Tracker with ReadMe File

Erix101:


TheCSharpAcademy:
@Erix101  Project approved! 😄✅ And well done for completing the challenges! Especially the support for multiple habits, great work!

🗺️And you're only one project away from getting your Olive Green Belt! 🫒🫒

🧑🏻‍🏫In terms of feedback, just a couple of things.

image

1 - Not only I don't see any point in creating a method that has just one line of code (why not just that one line of code when needed?), but also the recommended implementation is to just instantiate the connection when you need it. At the moment you create your connections without a 'using statement'. And when doing that you won't ever need to close the connection yourself. You can see good examples [in the first couple of code snippets here](https://learn.microsoft.com/en-us/answers/questions/804455/how-to-do-simple-crud-operation-on-sqlite-database): 

2 - I don't see the point of the times_used table. You don't want to have a table just for that. Maybe a config or settings table that will have times_used as one of the properties? Otherwise you're using a table as a property.

PR #276, opened 2025-03-08
Project Submission

Andrewki44:


TheCSharpAcademy:
@Andrewki44   Project approved! 😄✅ And well done for completing some of the challenges. Particularly the support of multiple habits using relational data 💪🏻

🥋And well done! You are one project away from getting the Olive Green Belt! 🫒🫒

🔍My only feedback is: Opening a connection in a singleton class is a bad idea. Don't worry about code repetition. It's almost always a better idea to create a short-lived connection inside each method where it's needed. 

image

PR #277, opened 2025-03-12
S1m0n3's Habit Tracker

S1m0n32002:


TheCSharpAcademy:
@S1m0n32002 Well done! Project approved! 😄🍾. And well done for completing the challenges! 👏🏻👏🏻👏🏻😎

🔍To be honest I don't have a lot of feedback, it's really good. The only thing I could be mention is that I was expecting habits that could be quantified (i.e. number of pages read, or number of chocolates eaten), but you demonstrated good command of SQL, which is the main objective of this project, so project approved regardless.😄

See you in the Coding Tracker. 🫡

S1m0n32002:
Thanks!
The Coding Tracker is also what i'm currently writing 😉

PR #286, opened 2025-04-02
Habit Tracker

KamilKolanowski:
Hi, here's my try on Habit Tracker, I hope everything's okay 👾👾


thomas-henley:
Hi @KamilKolanowski 👋

I'm going to take a look at your submission for the Habit Tracker. A lot of my comments will just be bits of advice to consider, not necessarily something you need to change. I'll annotate them for clarity:

🛑 : Something that needs to be fixed before the merge request can be approved.
🟨 : No fix needed, just something to think about. These will usually be coding style notes.
💚 : Something worth calling attention to for being particularly well done.

I'll comment here again when my review is complete.

HabitTracker.KamilKolanowski/Views/MainInterface.cs
```diff
@@ -0,0 +1,103 @@
...
+
+namespace HabitTracker.KamilKolanowski.Views;
+
+public class MainInterface
+{
+    public void Start()
+    {
+            DatabaseManager db = new();
+            HabitsPresentator hp = new();
+            var connection = db.OpenConnection();
```
thomas-henley, line 11: 🟨 : Usually we open and close a new connection for each database transaction. It's more scalable when you have multiple users hitting a server, but more importantly, it allows your application to recover from problems more easily. For instance, what happens if this connection breaks in the middle of the program running? There's no way to reconnect to the database. If you use a fresh connection each time, then the user can just retry the command if for some reason something goes wrong.

HabitTracker.KamilKolanowski/Views/PresentHabits.cs
```diff
@@ -0,0 +1,128 @@
...
+        Console.ResetColor();
+        Console.WriteLine();
+
+        int idWidth = 7;
+        int habitWidth = 20;
+        int quantityWidth = 12;
+        int unitWidth = 12;
+        int dateWidth = 26;
+    
+        string divider = $"├{new string('─', idWidth)}┼{new string('─', habitWidth)}┼{new string('─', quantityWidth)}┼{new string('─', unitWidth)}┼{new string('─', dateWidth)}┤";
```
thomas-henley, line 31: 🟢 : I can't believe you did all of this manually 😅 . It looks great! But I hope you've found Spectre since you did this, to save yourself a lot of work.

HabitTracker.KamilKolanowski/Views/PresentHabits.cs
```diff
@@ -0,0 +1,128 @@
...
+                string unit = i == 0 ? uom : new string(' ', uomWidth);
+                Console.WriteLine($"│{habit}│{time}│{unit}│");
+            }
+            Console.WriteLine(divider);
+        }
+    }
+
+}
+
+public static class StringExtensions
```
thomas-henley, line 101: 🟢 : Nice use of extension methods.

thomas-henley:
Excellent work @KamilKolanowski !

Feel free to comment on any review threads if you have questions.

🟢 : You went above and beyond, incorporating OOP and more advanced concepts like separation of concerns and extension methods. 

None of that was required for this project, but since you did it, I'll offer one piece of guidance at that level:

🟨 : Typically, you would not want your UI to directly interface with your database class. The DatabaseManager itself also prints to the console and gets user input. The distinction between the UI and database layer is pretty murky, with the UI calling database operations and the database manager interfacing with the user. There should be a service layer (and possibly a controller and/or repository class) between these two components, to reduce interdependency, separate responsibility, improve testing, and allow for greater modularity (for instance, if you wanted to change how you use your database down the line). None of that matters for this project though, and you will have ample opportunity to practice in future projects.

That's a minor quibble at this point, and a much more advanced topic than this project was meant for. You've done a great job on this. I'm especially impressed by the handcrafted menus!

✅ I'm approving this pull request. We don't actually merge them, so please close the PR when you can. Keep up the good work!

thomas-henley (approved):

HabitTracker.KamilKolanowski/Views/PresentHabits.cs
KamilKolanowski: yeah, this was pain in the ass, compared now to Spectre :D

PR #289, opened 2025-04-30
Habit Logger finished.

Golvi1124:
Tried to add as much as I wanted and could :)


src/Helpers/HelperMethods.cs
```diff
@@ -0,0 +1,234 @@
...
+    SELECT records.Id, records.Date, habits.Name AS HabitName, records.Quantity, habits.MeasurementUnit
+    FROM records
+    INNER JOIN habits ON records.HabitId = habits.Id";
+
+            using (SqliteDataReader reader = getCmd.ExecuteReader()) //(SqliteDataReader reader = command.ExecuteReader())
+            {
+                /*checking if the reader has rows and if it does we will read the data in a loop. 
+                For each row found in the table, we will add a new WalkingRecord to the list.*/
+
+                if (reader.HasRows) // check if there are any records in the table
```
chrisjamiecarter, line 36: 🟡 **Redundant**

💡 I would argue that `reader.HasRows` is redundant as `reader.Read()` will only fire if there is data (rows) to be read.

PR #299, opened 2025-06-14
Habit Tracker Implementation.

devjunaeid:
Used:
1. Database(Sqlite).
2. User Interface([Spectre.Console](https://spectreconsole.net/))
3. Database Interface(ADO .NET)


hasona23:
Hi @devjunaeid . How are you. Well done on the Habit tracker project. I really liked the design of the app you used SpectreConsole well .I just have a few suggestions 

1- Add a Measurement Unit to habits

2- Specify the time format needed to enter habit instead of having to look in source code to figure it out

3- Could a feature so that if entered "Now" it will get current date making it quick and convenient for user.

4- User Input could be put as a Helper Class instead of UI

5- I suggest using a Using statement when doing commands for sql like 
``` using var cmd = connection.CreateCommand``` to dispose of object at end of function/scope

6- try not to make the .sln file and .csproj it would make it much complex to handle diffent projects in the future like the ShiftsLogger where you will have to use 2 seperate projects

lastly the major error here it that there is no readme which is specified in the requirments of the project hence i can not accept however ill be glad to accept the project as soon as you add a Readme file

Thank you for the Good Project and happy coding 😄

**Review by hasona23 — CHANGES_REQUESTED** [review]
Missing Readme

**Comment by hasona23** [comment]
Hi @devjunaeid . I'm fine thanks for asking.

The README.md is really good. Well done on that one. and thanks for applying the suggestions.

I'm currently waiting till I can run the project and as soon as this is done I'll accept the project.

Regarding point 6 what i meant is a normal C# app would consist of 2 main parts 
1. the .sln file which is like the main file of the app 
2. 1 or more .csproj files so while making the app if using vs22 don't tick the option of put solution is same folder as .csproj.

so what I meant is that you make your architecture as follow
```
└── [ProjectName].[GitHubUserName]/
    ├── [SolutionDirectory1]/
    ├── [SolutionDirectory2]/
    └── [ProjectName].sln
```
Project name is you project HabitTracker then add your github name 
so ```HabitTracker.devjunaeid ```
then the .sln file is there with multiple solution directory. 
the solution directory is what contains the .csproj files for instance let's say you decided to separate code between a C#Library and the console app then you would make 
HabitTracker.Cli as sln directory 1
HabitTracker.Library as sln directory 2 
or any other name you want. this is what I meant by problems when expanding the project because in the future when doing larger projects like shiftsLogger you'll want to have 2 .csproj 1 for api(more on that later) and 1 for the Cli logic 

Is this ok  😊

devjunaeid:
Ohh I see, The issue was in the initial version the `data.sqlite` file does not contain the column 'unit' in the Habits table (That I've added later). I've added a **Reset** option to reset the DB with latest table definition.

PR #300, opened 2025-06-14
HabitLogger Project – Console App with ADO.NET and Report Features

Ana-Anna:


hasona23:
Hi @Ana-Anna . Well done in the Habit tracker project. The report feature is really good to see. anyway i just have a few suggestions.

-- Critical (cant accept project) -- 
1- ``` The users should be able to insert, delete, update and view their logged habit.``` 
I cant  delete the entries nor edit the habits and entries


-- Suggestions --
1- having a class called/file called helpers isnt much meaningful instead you could have Directory called Helpers then the helper files . moreover instead of helper.cs i noticed that this class is responsible for getting Input from user so could be InputHelper.cs

2- Instead of having parsing logic (which is part of input handling) in the HabitTracker file could add to the InputHelper.cs file mentioned in point 1 and have functions like GetDate() and GetName()

3-when deleting a habit i dont get the habits listed/displayed making it harder to choose instead will have to return to the ViewHabit command which is not a user friendly options

4- I suggest using (Spectre.Console)[https://spectreconsole.net/] it will make it easier to have a clean  good looking CLI and also make it easier to get input from user using for instance the selection option instead of having to enter ID manually and this will also allow you to hide the Id of the data if you want since user will not need it at this point if he will just select the habit he wants

lastly i must say the README.md is good and even included how to run the project more over the overall architecture of the app is really nice  using Dependency Injection  and having services classes which would be later in other projects so it is a good practice to do 

Really glad to review to the project.
Happy Coding 😄

hasona23 (changes requested):
changes required

Ana-Anna:
> changes required

Hi @hasona23 

Thank you for your thoughtful feedback — it really helped me improve the project.
I’ve implemented all the requested changes in a new pull request: [#322](https://github.com/TheCSharpAcademy/CodeReviews.Console.HabitTracker/pull/322).

Closing this PR to avoid duplicates and keep the review process clear. 
Thanks again for the support and for reviewing my project! 🙏

PR #301, opened 2025-06-17
Finished project

nibblenod:


TheCSharpAcademy:
@nibblenod Well done! Project approved! The code looks good and it fulfils all requirements. 😎👏

🔎My only feedback for you is to try to complete the challenges as well. In your professional career you'll stand out if you go beyond what's asked from you, even if just a little bit. Also, the challenges are a great way to reinforce what you've learned, since you have to add your own code to the tutorial's code. But nothing stops you from continuing to the next project and revisiting this one to extend it with the challenges.

See you in the next one!

PR #302, opened 2025-06-25
Finished project

kilozdazolik:
I only finished the parameterized queries challenge.


TheCSharpAcademy (changes requested):
@kilozdazolik Thanks for submitting! Your close! I saw you added some validation to the date, but it's still allowing impossible dates to be added, as per picture below. I suggest you use DateTime.TryParse() for your validation. Could you please fix before I approve? Upon fixing just commit and push your changes and your Pull Request will be updated.

image

NathanJenner:
@kilozdazolik - nice work on the application. One small comment regarding the app is using the term "Press any key to continue...". This doesn't actually work and you need to press Enter to progress. 
Users of applications can take things very literal at times.

For future code review please title your directory as follows:  projectname.kilozdazolik

PR #307, opened 2025-07-11
Project for review: Habit tracker.vagabundoo

vagabundoo:
Took feedback on last project on board with this one: https://github.com/TheCSharpAcademy/CodeReviews.Console.Drinks/pull/141


VocalNight:
Hello, nice job.
The program does everything it was asked to do, so i will be approving it.
That said, there are 3 things i would ask you to change and figure out for the next projects.

1- Find a way to clean the console after most operations, it can get quite messy if you don't.

image

2- Find a way to hide/format the date time when time is not needed, instead of leaving it as 00:00:00:

image

3- you left a Console.print somewhere in the code. This a minor thing, but make sure to take out all debug prints when you are doing a final code review

image

Overall, very good.
Keep up the good work!

PR #324, opened 2025-08-11
project finished

U-riani:


andymartinez1 (approved):
Hi @U-riani. Great job with this project! You have completed all of the requirements as well as the challenges. 

Just 2 things I would point out:

1. Why the count of records with no context at the top when first running the app?

image

2. The working directory should be moved. The project page shows how to do this: 
image

PR #326, opened 2025-08-22
HabitLogger.JJHH17

JJHH17:
This project allows users to:

- Add a habit (allows you to add a name, quantity, category and date of the habit.
- View a list of habits
- Delete all habits from the database
- Delete a habit from a given habit ID (which are all unique to the given habit).
- Edit a habit (from a given habit ID).
- Fetch the quantity of a habit, based on its habit category.
- Exit the program

It also seeds data into the database (adds 100 records per execution of the app) via a for loop.

### DIFF
--- a/HabitLogger.JJHH17/HabitLogger.JJHH17.sln
+++ b/HabitLogger.JJHH17/HabitLogger.JJHH17.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.14.36408.4 d17.14
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "HabitLogger.JJHH17", "HabitLogger.JJHH17\HabitLogger.JJHH17.csproj", "{42A28F90-941E-4665-A097-522E0379A081}"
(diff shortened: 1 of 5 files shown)


andymartinez1 (changes requested):
Hi @JJHH17. Excellent work on this project! You have completed almost every requirement and challenge. You will need to make a small adjustment to pass this requirement:
✅ Users need to be able to input the date of the occurrence of the habit

The "Add a habit" case has date validation: 
image

However, the "Edit a habit (by ID)" case does not:
image

Because of this, I can add anything for a date or not enter one at all:
image

Please fix that so I can approve this project. I also have some other feedback that isn't a requirement but may be helpful for future projects.

- As you mentioned in the README (great job on the README by the way), you should separate classes to have their own files within their own folders. For example, your class Habit should be in a separate file called Habit.cs within a Model folder. 
- Speaking of the Habit class, the naming convention for properties is PascalCase. For example, instead of ```public string name { get; set; }```, use ```public string Name { get; set; }```.
- Your seedData method (which should also use PascalCase instead) adds 100 new habits every time the program is run. I ran it 3 times and there are now 302 habits after I added the 2 from the screenshot above. Consider deleting old habits before seeding data or checking whether there is already data in the list. If there is data, there is no need to seed more data.

andymartinez1 (approved):
Excellent job @JJHH17. It now works flawlessly.

PR #332, opened 2025-09-19
Project with challenges

rafa-alves6:
# Parameterized queryes;
### Database seeding.


TheCSharpAcademy:
@rafa-alves6  Project approved! 😄✅ And well done for completing some of the challenges! 👏👏I hope you have learned a bit more with them. Seeding the database for example is very helpful in every project, not only for yourself as a dev to test things faster but also for us reviewers to review faster.

🥋And congratulations! You're only missing the Coding Tracker to get your olive green belt!! 🟩🟩

🔍️In terms of feedback, just one thing: To support multiple habits ideally you'd have a habits table and a records table instead of everything in a single table. Records would reference habits by its id to know what habit the record is for in a one-to-many relationship. You'll also learn how to separate your code in more classes and different files. You'll be using these concepts in the next projects. 👌

PR #355, opened 2025-10-16
Habit Logger

Esaiy:


TheCSharpAcademy:
@Esaiy  Project approved! 😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster.

🥋You're only one project away from getting your Olive green belt!! 🫒🫒 The coding tracker 👏

🔍️To be honest I don't have a lot of feedback, your code is clean and tidy and you follow good principles (SoC, DRY, etc). In the next project you'll start using more Object Oriented Program concepts so you'll work with more than one file.  Keep up the good work!

🔍One thing I could add is that for operations that require the user to input an Id, you want to show a list of records before asking for input, otherwise the users have to input something wrong to go back to the menu and find the id of the record they want.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

PR #363, opened 2025-10-29
NoNameLudovik.HabitLogger

NoNameLudovik:


NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Helper.cs
```diff
@@ -0,0 +1,54 @@
+﻿/*Helper - contains useful methods and classes*/
+
```
nwdorian, line 2: 🟡 **Code comments**

❓ It's fine to add comments during development if it's going to help you with pseudo-code or to remember some specific logic. When you finish a project, always go back through the code for a polish pass and clean up unnecessary comments.  

💭 Remember, our code should be self-documenting - with good naming and structure we shouldn't need any comments at all.  

➡️ Comments like this - explaining what a class does - are redundant, that should be obvious from the code inside the file.

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Helper.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Globalization;
+
+namespace NoNameLudovik.HabitLogger
+{
```
nwdorian, line 6: 🟠 **Namespaces**  

❓As per the academy's [code-conventions](https://thecsharpacademy.com/article/58/code-conventions), use file-scoped namespaces.  

➡️ From C# 10 we are able to remove the namespace block and save an indentation level, which optimizes the use of space and improves readability.

💡 If you're using Visual Studio 2022, here's a short article to setup [file-scoped namespaces by default](https://darthpedro.net/2024/01/29/configure-visual-studio-to-use-file-scoped-namespaces-on-new-classes/).
```suggestion
namespace NoNameLudovik.HabitLogger;
```

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Helper.cs
```diff
@@ -0,0 +1,54 @@
...
+            
+        }
+    }
+
+    public class DrinkWater
+    {
+        public int id;
+        public DateTime date;
+        public int quantity;
+    }
```
nwdorian, line 53: 🟠 **One class per file**

❓ Each top-level class, struct, record, or interface should be defined in its own file.

➡️ This practices keeps files organized and improves maintainability and discoverability. It also aligns with standard .NET conventions and matches the expectations of IDE tools and code analyzers.

nwdorian, line 53: 🟠 **Public fields**

❓Avoid exposing public fields directly. Use properties instead.

➡️ In C#, it’s generally recommended to use **properties** rather than public fields.

✅ Properties provide **encapsulation** - you can later add validation, logging, or computed behavior without breaking external code. 

✅ Properties are the standard convention in .NET for representing data members.
```suggestion
    public class DrinkWater
    {
        public int Id { get; set; }
        public DateTime Date { get; set; }
        public int Quantity { get; set; }
    }
```

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Menu.cs
```diff
@@ -0,0 +1,128 @@
...
+                
+                try
+                {
+                    Console.Clear();
+                    ShowRecords();
+                    Console.Write(@"Type in record's ID to delete it or 'b' to back to menu: ");
+                    int id;
+                    id = Helper.GetNumber();
+
+                    if (id == 0) break;
```
nwdorian, line 66: 🟡 **Logic bug**

❓ The message says enter `'b'` to exit but then you check for a `0` so pressing `'b'` doesn't work.

➡️ Make sure to align the display message with functionality.

```suggestion
                    Console.Write(@"Type in record's ID to delete it or '0' to back to menu: ");
                    int id;
                    id = Helper.GetNumber();

                    if (id == 0) break;
```

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+        static void Main(string[] args)
+        {
+            while (true)
+            {
+                Menu.ShowMenu();
+                SqlHelper.InitiateDataBase();
+
+                string? userInput = Console.ReadLine();
+
+                switch(userInput)
```
nwdorian, line 14: 🟠 **Magic Strings**

💡 You should use an Enum to avoid [magic strings](https://deviq.com/antipatterns/magic-strings). You can create a simple helper method like in the example below to show a human friendly version of the Enum in the menu.

Pasted image 20251026041947

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/Program.cs
nwdorian, line 1: 🟡 **Clean `Program.cs**

❓ Try and keep as little code as possible in `Program.cs` - it should only contain methods that initialize and run the project. 

➡️ Extract any logic into a separate method then simply call it from `Program.cs`

💡 In this case I would move this switch statement into the `ShowMenu()` method. You'd then be left with just two calls in the `Main()` method:

```
SqlHelper.InitiateDataBase();
Menu.ShowMenu();
```

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/SQLHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace NoNameLudovik.HabitLogger
+{
+    internal class SqlHelper
+    {
+        static string connectionString = @"Data Source=habit-Logger.db";
+
+        internal static void InitiateDataBase()
+        {
+            using (var connection = new SqliteConnection(connectionString))
+            {
```
nwdorian, line 15: 🟠 **Using blocks**

❓ C# 8 introduced a [using declarations](https://learn.microsoft.com/hr-hr/dotnet/csharp/language-reference/proposals/csharp-8.0/using#using-declaration) feature. We are able to remove a level of indentation by utilizing **using declarations** instead of using blocks and improve readability.

➡️ This feature, combined with file-scoped namespaces, removes a lot of indentation from the code and makes it much easier to digest.
```suggestion
            using var connection = new SqliteConnection(connectionString);
```

NoNameLudovik.HabitLogger/NoNameLudovik.HabitLogger/habit-Logger.db
nwdorian, line 1: 🟠 **Database in Repo**

💡 You shouldn't be committing your database file to source control. Add it to the `.gitignore`.

nwdorian (approved):
Hey @NoNameLudovik 👋,

Great work on your **Habit Logger** project submission! 🎉

I have finished the code review and left some feedback below. ☺️

Feel free to reach out if you have questions or if you'd like to collaborate on any of these points.✌️

**Emoji Legend**
- 🔴 Must fix issues 
- 🟠 Non-blocking issue to rectify going forward
- 🟡 Question, comment, observation, minor
- 🟢 Something I like

---

🟢 **Requirements**

⭐ You have completed all of the project requirements!

---

🟡 **Submission Structure**

❓ You have correctly added all the project files and folders into a top level folder `NoNameLudovik.HabitLogger` but `README.md` is outside of it. 

➡️ Include `README.md` in the top-level folder as well. 😄

---

🟠 **Project doesn't build**

❓ When I clone your project I get an error because we have different folder structure on our PCs. 

➡️ If there's anything that needs to be changed in order to run your project correctly add it in the README file. 

💭 I manually changed the path this time so it works, but you will be asked to change stuff like this in more advanced projects as reviews get more strict. 

---

🟡 **Validation**

❓ It's possible to add dates that are in the future. 

➡️ Add extra check for the date input being before `DateTime.Now`. This makes sense as we are recording habits from the past and it's a business rule. 

---

🟢 **Multiple files**

💪 Good work on separating logic into files. This will help you transition into OOP and it follows the [Separation of Concerns](https://thecsharpacademy.com/article/30005/separation-of-concerns-csharp) principle.

---

💡 You can take care of issues in your own repo if you want, but since the project will be approved you don't have to push the changes or create a new PR.

💭 I would love to see you attempt the extra challenges as it really helps in the long run, but they are not mandatory so it's up to you. 

💪 The application works well, it's easy to use, displays the information clearly and you've taken good care of error handling.

✅ I will go ahead and approve the project submission. Keep up the good work on the next ones! 😄

Best regards,  
@nwdorian

NoNameLudovik:
Hi @nwdorian , thanks for your feedback. I’m trying to process your comments, and I have a question about the issue with building my project. Could you please give me more details about it? 
Sorry if I'm bothering you.😅

nwdorian:
@NoNameLudovik Hi 🙂 

You are not bothering me, when I say reach out if you have any questions I mean it 😄 

I forgot to add a picture of the error in my review that might have made I clearer, here: 

image

This is because of of the `workingDirectory` field in `launchSettings.json`
- `"D:\\LearnC#\\Repositories\\CodeReviews.Console.HabitTracker\\NoNameLudovik.HabitLogger\\NoNameLudovik.HabitLogger\\"`

That path is specific to your PC - anyone else that clones the project will have a different folder structure. 

I see this in a lot of Habit Logger projects and I believe it's because of the "Changing Your Working Directory" section on the project page. I'm not sure why this is happening, if I recall correctly - before it would just add a relative path to the project root instead of the full path of the machine. 

Simple way to tackle this is to just change the path to `"."` so it uses the root of the project. 

```
{
  "profiles": {
    "NoNameLudovik.HabitLogger": {
      "commandName": "Project",
      "workingDirectory": "."
    }
  }
}
```

This kind of thing is quite intricate to test as "it works on your machine" so you have no issues, but these are the joys of software development. 😄 Once you switch to hosted database like SQL Server this won't be a thing.

PR #366, opened 2025-11-13
HabitTracker

smad2:
I think the Habit Analytics section is a good challenge to add to this project.


smad2:
I can't make the changes that Codacy tells, otherwise my app will break

TheCSharpAcademy:
@smad2  Project approved! 😄✅ And well done for completing the challenges! This is one of the best, if not the best habit tracker I've ever seen and this is a project present from the very beginning of The C# Academy when there were just 3 projects. Well done! 👏👏

🥋You're only one project away from getting your Olive green belt!! 🫒🫒The Coding tracker 🧑‍💻

🔍️**Feedback**
I'm very impressed with the UI and the reports. I've never seen such details reports in any console apps. That's a truly great standard. Your code is also clean and tidy and follows good OOP principles,  SoC and DRY, which is fantastic. I think the only think I tend to stay away from is wrapping one-liners in a method (I prefer to just call the one-liners directly), but in your case it makes sense, since your set of UI wrappers is pretty consistent:

💡**Suggestion:**
When I created the Academy I was still an entry-level dev. The more I progressed in my career the more I understood the importance of unit tests. You can see the Unit Tests project is at the end of the console area. But if I was creating the academy again today I'd put it right at the beginning since unit tests are super helpful not only to write better code overall but to make it less buggy. 

So my suggestion is that you skip everything else and [complete the Tests project](https://thecsharpacademy.com/project/21/unit-testing) and start using unit/integration test from now on in ALL of your projects. You don't need to fully cover your code with tests, but I think it's good to develop this habit from very early on. Feel free to reach out on Discord if you have any questions!

PR #369, opened 2025-11-21
HabitTracker

gl-trapani:


andymartinez1 (approved):
Hi @gl-trapani. Excellent job on this project! You did a great job with Separation of Concerns. All classes make sense and are organized very well. The README file is well done too. Not the generic AI generated stuff we typically see.

My recommendations:

1. Try adding a unit of measurement when adding a habit. Ex: 

image

It would make more sense saying something like "Enter cups of water" to track the actual amount. Same goes for exercise and homework. For exercise, it can be something specific such as pushups or situps.

2.  I would always recommend attempting all the challenges:

Especially seeding data. That will be important soon
image

But overall, you did a very good job! I look forward to seeing your next submissions.

PR #371, opened 2025-12-01
My Attempt to the Habit Tracker project!

diegopetrola:
Hello, this is my attempt at the Habit Tracker project!

The README file should have enough information to run and understand the basic idea of the project. 

After grading, please go to `C:\Users\<USERNAME>\AppData\Local` and delete the file named `habit.db`. In Visual Studio it's a hassle to create files dynamically at the top level of the project because it's binaries run in another folder, so I had to do that instead.

Full disclosure: I skipped the "do not use EntityFramework" part. I already have a good grasp of SQL and understand the security importance of parametrized queries, so I opted to use this opportunity to familiarize myself with dotnet specific tools. I hope the fact that I did all the challenges can make up for that.


andymartinez1 (changes requested):
Hi @diegopetrola. When following the README, I get this error message:
image

With Entity Framework, the user will need to "Update Database" in order to run this. Make sure to include that in the README. 

The issue of where the .db is located can be solved by changing the working directory:
image
This is included in the project page:
https://www.thecsharpacademy.com/project/12/habit-logger

Also, it is bad practice to have negative number IDs. It can be a bit confusing.

We can let this requirement slide if you're comfortable with SQL:
image

But this requirement will need to be addressed for me to approve it:
image

While there is an option to view, create and delete the habits, there is no way to update them. Please update this so I can approve. I look forward to seeing the finished project.

diegopetrola:
I have addressed all your concerns, namely:

- Added option to edit both Habit Types/Caterogies and Habit logs (how did I forget that?)
- The project now runs inside the current folder and the database is there (very valuable info, I googled a little and every information I got was either overly complicated or not elegant like typing '../../' inside my code...)
- Removed negative Ids
- The `Update-Database` is not needed now because the db is inside the project and already configured, but I added extra info on the README to explain how to install and run the code

Thank you for your review, andymartinez!

andymartinez1 (approved):
Hi @diegopetrola. Great job with the changes! I am approving the project.

PR #377, opened 2025-12-09
Finished Project

AdamWayne-dev:


TheCSharpAcademy:
@AdamWayne-dev Project approved! 😄✅ And well done for completing the challenges! You implemented the support for multiple habits perfectly. And I really like the "bad habits" twist 👏👏

🥸**Feedback**: 
🔍️Although code organization is not a major concern at this stage, i'd like to point out that some of your methods are doing too many things: Collecting input, querying the DB, displaying messages, etc. In the future think of how you could split your methods so your code is more modular and hence more testable. 

🔍️If you're going to measure habits by time ideally you should add validation to avoid this type of monstrosity: 

image

🔍️For the next projects, consider adding unit testing, it's a very important skill and the earlier you learn it the better. I'm slowly including it as a requirement or challenge in more projects, [you can find an article on unit tests for beginners here](https://thecsharpacademy.com/article/30009/unit-testing-for-beginners). 

PR #379, opened 2025-12-12
Finish Habit Logger

coolorvi:


TheCSharpAcademy (changes requested):
@coolorvi Thanks for submitting!

At the moment I'm not able to submit a habit record, could you please let me know what I'm doing wrong? Ideally you'd want to warn the user with the exact error their input has 😢. 

image

coolorvi:
Thanks for review. I have fixed some errors related to checking the correctness of input.

TheCSharpAcademy (changes requested):
@coolorvi Well done! The project is looking great. There's a big problem though, and although it's part of the challenges and not the requirements I can't approve since it's a really bad mistake 😔:

To support multiple habits **you shouldn't allow users to create tables**. You need to create two tables only (habits and records), joined by a foreign key, in what's called a **one-to-many relationship**. It's important that you understand this concept right away before moving forward, so could you please make this change before I can approve?

coolorvi:
I have completely rewritten the project. Now only two related tables are created.

TheCSharpAcademy:
@coolorvi  Project approved! 😄✅ Well done for reworking your project. Usually we don't review based on the challenges but since this is such an important concept it was better to prevent you from moving forward without understanding this key concept of database design🤓

🥋You're only one project away from getting your Olive Green belt!! 🫒🫒 The Coding Tracker 👏. In this next project you'll dive into Object Oriented Programming and you'll use Spectre Console for better management of UI in console projects. Are you ready? 😎

👪If possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

PR #380, opened 2025-12-18
Habit Logger Project completed

aneevel:
I won't repeat what I included in my README here, that should provide any needed information :)


TheCSharpAcademy:
@aneevel  Project approved! 😄✅ Well done in completing some of the challenges, particularly Unit Testing. That's a great habit to create very early in your journey 👏👏

🔍Feedback
1️⃣I can see you took your time to try different things in this project, and it feels a bit over-engineered (meaning you could have solved it with way less code), but that's ok. We do encourage you try different approaches to practice areas you feel you need, instead of just ticking requirement lists thoughtlessly. 

2️⃣The one thing that jumps out is that your "Engine" seems to be doing too much. For starters, ideally you want to name it a bit more clearly. What Engine? What does it do? It sounds a bit vague. Also, all of your GetInput methods are in your engine (as per picture below). Seems to be like a violation of Separation of Concerns. 

image

3️⃣Also I'm not sure about the need for multiple projects (other than of course the tests project). I'd like to have seen an explanation for that architectural choice in your Readme 😊

👨‍🏫The above are things to thing about, but overall great work 👏👏

PR #381, opened 2025-12-23
Review

filipmarkovic064:
Tried doing everything except for unit tests, i will do them in the next project.


TheCSharpAcademy (changes requested):
@filipmarkovic064 Well done! The project is looking good. However I'd like to ask you to rework the way you solved the "Multiple Habits" functionality. Ideally you want to have two tables: Habits and Records, linked by a foreign key, in what we call a "one-to-many" relationship.

I know this is only part of the challenges, but it's important that you understand this concept sooner rather than later 😊. Let me know if you have questions!

filipmarkovic064:
Hey thanks for the review! I totally get what you meant, i implemented what we learned in Flashcards project and made 2 tables now. 
I had to rewrite most of my functions to implement both functions and my code was an eyesore (all in 1 class) so i made a few classes, its mostly the same logic except that its based on 2 tables (Records and Habits) and that i moved methods around so that its easier to navigate. Let me know if theres anything else i should fix

TheCSharpAcademy:
@filipmarkovic064 Thanks for the changes! Project approved! 😄✅ And well done for completing some of the challenges! The table relationship is super important and you nailed it 👌

**Feedback**:
🔍Some of methods at the moment do too many things. Your ChooseHabit() for example: 1. Reads habits from the DB, 2. Prints them, 3. Prompts Input, 4. Collects input, 5. Shows result. It's ok to have an "Orchestrator" method that is responsible for tying a bunch of actions together, but each of these actions could be outsourced to it's own method, each with a single responsability. For this reason I highly recommend you jump into unit tests for the next project. You don't need to cover the whole project with tests, but at least writing a couple of tests will give you an idea of how to better break your methods down, which makes them more testable. [You can find an Intro to Unit Testing here.](https://thecsharpacademy.com/article/30009/unit-testing-for-beginners)

🧼Please please please clean up before committing. Each file in your project starts with 10 lines of unused usings (aka garbage).  [You can read our code conventions here](https://www.thecsharpacademy.com/article/58/code-conventions)🤓.

image

🔍I suggest you look into **top level statements**. Your Program.cs could have 2 lines instead of 19:

image

✋🏻Overall great work! 😎Looking forward to seeing your next projects! 🫡

PR #382, opened 2025-12-23
project finished

gjr1440:


TheCSharpAcademy:
@gjr1440  Project approved! 😄✅ And well done for completing the challenges! One of the most important challenges at this stage is the creation of a one-to-many relationship in this project and you implemented it perfectly 👌👌. For the next one I highly recommend you also complete the unit tests challenge. It will take your dev skills to the next level. 

🥋You're only one project away from getting your Olive Green belt!! 🫒🫒 The Coding Tracker 👏 Get ready for a much more elegant approach for embedding SQL in C# code 😊

🔍️In terms of feedback, your code is super clean and well organized. There are just two things I can mention. I think the names of your classes are very generic and it's not quite clear what's happening. I'd swap all of these names: 

- Database = DatabaseInitializer
- Helpers = InputHelpers (here I'd split the validation into it's own static methods so they're more easily testable)
- Habits = HabitsService. List = GetHabits. Insert = InsertHabit. 
- Records = RecordsService.

I think you get the idea. Don't be afraid to be verbose when naming stuff. You want to leave no margin for doubt. 

🔍Also, when you list your records, I can't see the habit name or unit of measurement, only the Id, which is not very useful to the user (see picture below). With a JOIN query you could retrieve records with the habit name and unit of measurement. I suggest you try this even after having this project approved. JOIN queries are extremely important and the sooner you learn them the better. 

🍎Besides the feedback above, overall great work 👏👏

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 

PR #384, opened 2026-01-03
HabitLogger submission

ghost:
Thank you :)


andymartinez1 (approved):
Hi @m1chael888. Excellent job on this project! Everything seems to be working as intended. The README was very detailed and showed your thought process. 

I will approve this project but here are a couple of things to consider for future projects:
1. The README didn't contain any instruction on changing the launchSettings.json working directory to get it working. I would consider not using a launchSettings.json at all for a console app. Let it use the root directory so it can start working on any machine right away.

2. There are a few issues picked up by Codacy. Make sure you look at this when submitting a project as it tells you if there are unnecessary using statements or unused variables as shown in the screenshot below:
 
image

PR #393, opened 2026-02-05
project finished

slimeswamps:


andymartinez1 (changes requested):
Hi @slimeswamps. Your project is looking good so far. However, you are missing this requirement:

Image
<hr>

At the moment, it just takes today's date even when updating the record. Please fix this so I can continue with the project review.

andymartinez1 (approved):
Hi @slimeswamps. Great job with the fix provided. I am approving the project as all requirements are now met, but I will provide some additional feedback.

<hr>

⚠️ For DateTime format:

I would suggest using (YYYY-MM-DD) since this is the international standard for representing dates and times (ISO 8601).

<hr>

Other Feedback:
⚠️ As your project currently is, you have to enter after getting an error but there is no indication that you have to press enter to continue. I would recommend adding a prompt that says you need to hit enter to continue. 

<hr>

Overall, you did a great job! I like your approach to the menu. I look forward to reviewing your future projects!

Best regards,
@andymartinez1

PR #396, opened 2026-02-18
Habit Tracker project

natholder:
habit tracker done wahoo!


andymartinez1 (changes requested):
Hi @natholder. Great job with this project! Everything is looking good except for one issue:
When entering a value in the main menu that isn't 1-5, it closes the program. Instead of the default case exiting the program, you should have case "5" exit and the default to warn the user that the input is not valid. Please fix this issue so I can approve the project. 

Other feedback

<hr>

⚠️ For DateTime format:

I would suggest using (YYYY-MM-DD) since this is the international standard for representing dates and times (ISO 8601).

<hr>

✅ The README was well thought out and included our requirements:
- Project description
- How to build and run the project 
- Thought process behind the project (What was hard? What was easy? What have you learned?)

<hr>
⚠️ You should validate the entry first and then ask for new values only if the ID exists. As it is now, it asks for values before even checking if the ID is in the db

Image

<hr>

Other than that, everything is looking good. No unused imports or inconsistent code/formatting. Once you make that quick change to the main menu options, I will approve the project.

Best regards,
@andymartinez1

andymartinez1 (approved):
Hi @natholder. Great job with the project! You shouldn't put a switch statement in an if-else block. The default case takes care of that. 

<hr>

I am approving this project. I look forward to reviewing your future projects!

Best regards,
@andymartinez1

PR #398, opened 2026-03-01
Finished and submit Habit Tracker project for review

pillipillow:
Hi. This one took a while. Hopefully what I've coded here is suffice and achieved most of the requirements and challenges. Otherwise, please let me know what I did wrong.


TheCSharpAcademy:
@pillipillow  Apologies for the delay in the review. Project approved! 😄✅ I love that you completed some of the challenges!  And really thoughtful Readme! Well done!

**Feedback**
🔍️I highly recommend you complete the challenge to use parameterized queries. Inserting data via string interpolation opens your application to SQL injection attacks. For example, a skilled hacker could easily create an input to drop a table. 

🔍Your foreign key is declared, but there is no cascade delete defined. This means you'll have orphaned records. The objective of foreign keys is to keep data integrity. Using a "ON DELETE CASCADE" would ensure all records for that habit would be deleted as well. Alternatively if you don't want that behavior, you can use what we call a "soft delete", where the habits aren't actually delete, but they are just marked as deleted. I'll let you do your own research on this. (Really good question by the way)

🔍In this project it doesn't make sense to let the users add a future date. Also it shouldn't allow negative quantities (see picture below). I recommend you always think carefully about what types of inputs should be accepted:
<br>
image

🧑‍🏫Overall well done! 👏👏 Looking forward to seeing your next projects!

PR #399, opened 2026-03-03
Project finished

CSamu7:


TheCSharpAcademy (changes requested):
@CSamu7 Thanks for submitting! We're really sorry about the delay in the review 😟. The project is looking good but unfortunately there's a big bug in your project, which is stopping it from metting the requirements: at the moment updating doesn't work. It only works with the date, and only if the date is changed using "today". Could you please fix before we can approve!

TheCSharpAcademy:
@CSamu7  It's been two weeks since the suggested changes, so we're archiving this project. Feel free to reopen the PR and resubmit in the academy when you're ready. 🤓

PR #400, opened 2026-03-03
Request for review of project

Kerem1989:


TheCSharpAcademy (changes requested):
@Kerem1989 Thanks for submitting! Apologies for the delay in the review 😟. There's a major bug in the app so I can't approve it. As you can see in the picture below, at the moment when I input a date it fails silently and just replaces it with the current date. Users need to know what's the expected format and if there's an error they we need to show a proper error message.

Image
<br>
Also, every time the app starts it drops the database, which defeats the purpose of the database altogether. 
<br>
Could you please fix so we can approve?

TheCSharpAcademy:
@Kerem1989   It's been two weeks since the suggested changes, so we're archiving this project. Feel free to reopen the PR and resubmit in the academy when you're ready. 🤓

PR #403, opened 2026-03-10
Project Submitted

kedwards1306:


TheCSharpAcademy:
@kedwards1306  Project approved! Apologies for the delay!😄✅ And well done for completing the challenges! I recommend you always do that, as it pushes you and makes you learn faster. And thanks for making a nice readme with your thoughts. It's super helpful for the reviewer 👏👏

🔍️In terms of feedback, just a couple of things: First, when using a foreign key you don't need to delete the children of a record manually as you did in the picture below. If you use "ON DELETE CASCADE", Sql will do that for you and guarantee data integrity.<br>
image

🔍️Looks like when updating a record the ids were changing. It was very hard to find the modified record after updating it. 😟. It seemed like the logs were randomly ordered. I suggest you order it by date or id next time.

Overall good work!✋🏻Looking forward to seeing your next projects!

PR #404, opened 2026-03-16
PEx Habit Tracker

ProximaEx:


TheCSharpAcademy:
@ProximaEx  Project approved! 😄✅ And well done for completing the challenges! Your project is very creative and has a lot of personality. I love the animations and the UI in general. Oh that's not to mention the very thoughtful Readme. It's amongst the best I've seen. Feflecting upon what you've learned the way you did takes your learning to the next level. Keep up those standards!

🔍️I strongly recommend you look into parameterized queries. It's not good practice to use interpolation in your SQL queries as it makes your code vulnerable to SQL injection. SqliteConnection has a lifecycle: create it; open it; use it; close/dispose it; stop touching that instance. For a small project without a lot of traffic, you might not experience issues but for real world usage, using a single instance will cause weird behaviors.

🔍️You're opening and closing your connection in every request and that's a positive, but the best practice is to create a connection instance in every request, so disposal is clean and predictable. In C# a disposed object is meant to be finished with permanently. Not reused. 

🔍️In supporting multiple habits you've created two tables and that's correct. But you didn't use these tables correctly. To associate a habit to a record you should use the habitId, not the habit name. The habitId would be a FOREIGN KEY which would link these tables. I suggest you rewrite this before moving forward, that's a crucial concept to understand.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

Overall great work!✋🏻Looking forward to seeing your next projects!

PR #408, opened 2026-04-08
finished

nice-Code-90:


TheCSharpAcademy (changes requested):
@nice-Code-90 Thanks for submitting! At the moment your project doesn't compile. Could you please fix? 

Image

nice-Code-90:
> @nice-Code-90 Thanks for submitting! At the moment your project doesn't compile. Could you please fix?
> 
> Image


I've fixed it, thank you for review!

TheCSharpAcademy:
@nice-Code-90  Project approved! 😄✅ And well done for completing some of the challenges! I recommend you always do that, as it pushes you and makes you learn faster. I'm very glad you added a test project. Great habit to create early on! And you even added Azure Speech! Did it work well?

📰Also, thanks heaps for your Readme! I suggest you always finish your projects with reflections like this. Not only it helps people see how you think but it's also useful for your learning. It cements your knowledge and helps keep your motivation. Yeah it's completely different to work on an established code base and having to make design decisions yourself, even for small projects! Don't worry, you'll have to deal with a bunch of Senior level decisions throughout the roadsmap. 😎

🥋You're only two projects away from getting your Olive green belt!! 🫒🫒 The Habit tracker and the Coding Tracker 👏  It's time to start mixing SQL and C# together. Are you ready? 😊

**Feedback**
🔍️At the moment I can enter an empty unit of measurement. It would be good to add NullorEmpty to your validation. See below:

image<br><br>

🔍️It would be good to let the users change individual properties when updating, as opposed to asking them to update every property. More often than not updating doesn't involve all properties, which is tantamount to deleting the old property and adding a new one. Alternatively, if prompting all properties, at least give the users a binary option (keep current value?), etc. 

🔍️It's very good you added integration tests but equally important would be to test some of the business logic with unit tests. A great candidate would be the validation of data. Great way to tests many weird inputs (such as empty unit of measurement). 

🔍️To support multiple habits it would be better to have two tables linked by a Foreign Key to enforce data integrity. At the moment I can have, for example as habits: Running, Runnning, runing, running, etc. A single Running habit would ensure that all running records would use the same running entity and properties.

Overall great work!✋🏻Looking forward to seeing your next projects!

PR #410, opened 2026-04-23
Finished the project

Sephydev:
Finished the project !

All the challenges I had is explained in the README.md. The only comment I want to add is about the typo in the name of the project. I've only noticed the typo ("habbit" instead of "habit") at the end of the project, and I'm quite scared of the headache it could be to change it and refactor everything linked to it. So I prefer to leave it like that and focus on the next project !


TheCSharpAcademy:
@Sephydev  Project approved! 😄✅ And well done for completing the challenges! I love that you're starting unit tests this early in your journey. That's a great habit to create, the earlier the better! 👏👏 

🔍️One thing I would have done differently, but it's not a blocker since you'll learn it anyways down the road: To support multiple habits it would be better to have two tables: Habits and HabitRecords, linked by a foreign key.

🔍It's not wrong to use tuples like this but it would be much better to have a custom type for better encapsulation and readability. You could group all of these properties under a HabitRecord type. This way you can pass this around without having to show that gigantic monster everywhere:

image

PR #411, opened 2026-04-28
Initial commit

urasylmaz1:


DSills735 (approved):
Hello @urasylmaz1 , I am here to review your code! It looks like you have done great work on your proejct. 

It is a shame that you did not do the challenges. At a minimum, I would HIGHLY reccomend to write some unit tests and go through that process. That came back to bite me later in the roadmap when I ignored that challenge on this project. 

I am going to approve your project, but I do have some considerations to get your mind moving for future projects:

**String Interpolation in SQL Requests:**

This is where SQL Injection happens. You might consider another way to do this, as It does pose a security risk, even though it works. 

**Try Catch Blocks**

Around your DB calls, consider using Try Catch blocks in case the DB fails for some reason. This becomes more important when you are using a hosted DB, or something like SQL Server.

PR #413, opened 2026-05-07
feat: first version Habbit Logger

SylvesterNdwata:
All feedback is welcomed.
Thank you in advance.


DSills735 (approved):
Hey, @SylvesterNdwata,  Great work on your submission, I will approve! 

I highly recommend you take a look at the unit testing challenge, and consider doing it on your own time. Eventually unit testing will be required. It is very good info to learn early, and fairly simple to start out! 

I also want to stress to use Console.Clear whenver you can to keep the console as simple as possible. Its not a critical fail at any juncture, just helps clean some messy UI. You have it implemented intermittently. The reason why I recomnend this now, Is because soon you will have some more expansive projects, where the UI is much more robust.

Nevertheless, great work on the project. See you on the next one!

- @DSills735

